PR #1, opened 2017-01-24
Mvp

manolo:
Fixes https://github.com/vaadin/components-team-tasks/issues/228

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-input/1)


limonte:
:lgtm:

---

Review status: 15 of 16 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/1#-:-KbKtgGOyvnsnk3WHiEx:bnfp4nl)*

manolo:
Review status: 15 of 16 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[vaadin-input.html, line 47 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/1#-KbFNWVL3l06MQdgxlT7:-KbKuRg99Q1Y4QkE3RR1:b-m6gnmu) ([raw file](https://github.com/vaadin/vaadin-input/blob/60317c6ecd88e5e52e77e4b98a0a60ad94bc38ce/vaadin-input.html#L47)):*

Should we turn all the native attributes to use attributes? `paper-input` does that even for `name`, `list`, `pattern` and `placeholder` and `title`.
</blockquote></details>

I'd rather not having those values in the dom if not absolutly necessary.

---

*[vaadin-input.html, line 233 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/1#-KbJgPl9sLLGvnfJknkZ:-KbKu8qC8sL3Wbettmtm:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/fb99f0c8dfe55dd431678073f968266382393a6e/vaadin-input.html#L233)):*

As discussed in chat, apparently is not needed right now although iron-input does it. I will remove it until we observe a related issue.
</blockquote></details>

Done.

---

*[wct.conf.js, line 10 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-input/1#-KbKWE2GSujT2I_2VgC9:-KbKtuclTQYbzTeoMSx-:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/890247b83019553c43308485c78353c919af0b18/wct.conf.js#L10)):*

Should we test on `iPad` as well?
</blockquote></details>

Done.

---

*[test/validate.html, line 16 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-input/1#-KbKI9jApgLGTiCsntQ5:-KbKuHPOYqNWN8aw4qDk:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/890247b83019553c43308485c78353c919af0b18/test/validate.html#L16)):*

Might make sense
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/1)*

Saulis:
Review status: 11 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-input.html, line 173 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-input/1#-KbKv8Px_i5FeckoGac2:-KbKv8Px_i5FeckoGac3:b2dxls6) ([raw file](https://github.com/vaadin/vaadin-input/blob/b6ea2d2e9c410ed83834ff8f96e6a47d9cb8c3f5/vaadin-input.html#L173)):*
> ```HTML
>              * https://developer.mozilla.org/en/docs/Web/HTML/Element/input#attr-type
>              */
>             type: {
> ```

We shouldn't have `type` at this point at least.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/1)*

manolo:
Review status: 11 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-input.html, line 173 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-input/1#-KbKv8Px_i5FeckoGac2:-KbKvl__Rz8T_JPqBlr2:b-skt6uk) ([raw file](https://github.com/vaadin/vaadin-input/blob/b6ea2d2e9c410ed83834ff8f96e6a47d9cb8c3f5/vaadin-input.html#L173)):*

We shouldn't have `type` at this point at least.
</blockquote></details>

How do you define a password input then ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/1)*

platosha:
Reviewed 1 of 10 files at r2, 1 of 1 files at r4, 2 of 3 files at r5, 4 of 5 files at r6.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/1)*

Saulis:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-input.html, line 173 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-input/1#-KbKv8Px_i5FeckoGac2:-KbOkDSDpBXy0ntgdN-N:bu4mseo) ([raw file](https://github.com/vaadin/vaadin-input/blob/b6ea2d2e9c410ed83834ff8f96e6a47d9cb8c3f5/vaadin-input.html#L173)):*

How do you define a password input then ?
</blockquote></details>

That question will be answered when we will start working on the "custom validators" and "prevent invalid input" tickets. What we do know for sure is that we don't want people to expect that something like `type="checkbox"` or `type="button"` would work, hence it's better to remove the property for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/1)*

manolo:
Review status: 13 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-input.html, line 173 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-input/1#-KbKv8Px_i5FeckoGac2:-KbPGCbOyRRKKMIWerfU:blyrpa4) ([raw file](https://github.com/vaadin/vaadin-input/blob/b6ea2d2e9c410ed83834ff8f96e6a47d9cb8c3f5/vaadin-input.html#L173)):*

That question will be answered when we will start working on the "custom validators" and "prevent invalid input" tickets. What we do know for sure is that we don't want people to expect that something like `type="checkbox"` or `type="button"` would work, hence it's better to remove the property for now.
</blockquote></details>

Thinking more about this, I feel like we could have  a common vaadin-native-input or whatever name,  with the `type` and all common stuff, then  all other vaadin form controls could wrap or extend it exposing only the native input api that we want for it.  For instance for `vaadin-input` we support text/password/url/email ... via attributes or something.

Anyway removing the type for now.

Done. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/1)*

Saulis:
:lgtm:

---

Reviewed 7 of 18 files at r1, 4 of 10 files at r2, 1 of 1 files at r4, 1 of 3 files at r5, 3 of 5 files at r6, 3 of 3 files at r7.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/1#-:-KbPNgzQ-yzMkjEwr6r5:bnfp4nl)*

Issue #9, opened 2017-01-29
autofocus doesn't work in Linux/Firefox


limonte:
![](https://raw.githubusercontent.com/vaadin/vaadin-input/54a1230ce7c560e4bf3ef430b3ce06ca5d1aae96/test/visual/screens/vaadin-input/default/default/firefox.png)

Related to #6

limonte:
http://stackoverflow.com/a/9956938/1331425

> Autofocus is only done before onload has fired; it's meant to be a declarative way of specifying focus on initial page load.

manolo:
Is there anything that we want to do with this?  Or a nice solution ?

manolo:
I meant without recurring to JS.  Anyway sent a PR that fixes this with a bit of JS.

PR #17, opened 2017-02-14
Fix tabindex

manolo:
Implementation of control-state in a separated mixin that could be used in any other custom element.

It works in shadow and shady dom, and handles correctly tab and shift-tab.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-input/17)


platosha:
Reviewed 7 of 7 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 14 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw-8a4NXsJsLTuJmM9:-Kcw-8a4NXsJsLTuJmMA:bvnxv0m) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L14)):*
> ```HTML
>           focused: {
>             type: Boolean,
>             value: false,
> ```

Perhaps it is unnecessary to specify the `false` value.

If you do specify, though, then `disabled` should also have it for consistency.

---

*[vaadin-control-state-mixin.html, line 36 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw1Hf0d-vOqUTGaHtm:-Kcw1Hf0d-vOqUTGaHtn:b-li7wb9) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L36)):*
> ```HTML
>            * @private
>            */
>           tabindex: {
> ```

We can’t use `attributeChangedCallback`/`observedAttributes` for that, right?

Maybe document that developers can use `tabindex` attribute instead of declaring as “Internal” and “@private” then?

---

*[vaadin-control-state-mixin.html, line 56 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw4MS527DucwRDF16Z:-Kcw4MS6A0kGH4nL-Q-q:b-82y97n) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L56)):*
> ```HTML
> 
>       Polymer.RenderStatus.afterNextRender(this, () => {
>         this._getFocusElement().addEventListener('focus', this._onElemFocus.bind(this));
> ```

if the implementer element happens to define `_onElemFocus` method as well, will the implementer method used here then?

If so, we should maybe find a better way of protecting these event handler methods from accidental override.

---

*[vaadin-control-state-mixin.html, line 62 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw5UZmgVZmjQeoQ_pq:-Kcw5UZmgVZmjQeoQ_pr:b-4pj8i6) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L62)):*
> ```HTML
>     }
> 
>     _getFocusElement() {
> ```

If this is a mixin’s API, it should be documented.

Btw, maybe a getter method?

---

*[vaadin-control-state-mixin.html, line 49 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7jGmtqkZmcFQkgPq:-Kcw7jGmtqkZmcFQkgPr:b-j2km35) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L49)):*
> ```HTML
>       // Make FF take the focus (#9)
>       if (this.autofocus) {
>         this.focus();
> ```

This relies on the implementer element to implement the `focus()` method. 

Maybe give non-implemented warning in console?

Alternatively, making `_focus()` public, as suggested below, will also resolve this.

---

*[vaadin-control-state-mixin.html, line 71 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7Txh5hR1AaJAlbVW:-Kcw7Txh5hR1AaJAlbVX:b-66rwsp) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L71)):*
> ```HTML
>     }
> 
>     _focus() {
> ```

Maybe expose `focus()` and `blur()` as public right away from here?

---

*[test/control-state.html, line 31 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw6mNxacgZ6Qp6W8Gf:-Kcw6mNxacgZ6Qp6W8Gg:b-8mitwc) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/test/control-state.html#L31)):*
> ```HTML
> 
>       describe('tabindex', function() {
>         if (!window.ShadyDOM) {
> ```

Let us combine this test with the other shady-DOM soped tests below in a single `if ... else ... `. This way would be easier to see what is different.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

tomivirkki:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 5 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw2d65ioE_5zTaAil5:-Kcw2d66g3WFyYvZNOh1:b-5638uc) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L5)):*
> ```HTML
>   // IronFormElementBehavior is not a proper 2.0 class, and its tabindex
>   // implementation fails in the shadow dom.
>   window.VaadinControlStateMixin = subclass => class extends subclass {
> ```

Ouch, I didn't realize we're polluting the global namespace already. Our stuff should probably be under `vaadin` namespace?

---

*[vaadin-control-state-mixin.html, line 67 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw5CXcLFZCOfQM9gD4:-Kcw5CXcLFZCOfQM9gD5:bh90wuf) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L67)):*
> ```HTML
>     }
> 
>     _onElemFocus() {
> ```

Why "Elem" instead of "Element"? Let's use "Element" everywhere for consistency.

---

*[test/control-state.html, line 73 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-KcwBOzaJ26Ih4K9OUrP:-KcwBOzaJ26Ih4K9OUrQ:b-hbkftg) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/test/control-state.html#L73)):*
> ```HTML
>         it('should focus the internal input on focus', function(done) {
>           // Avoid FF window focus issues by stubing native focus behavior
>           window.sinon.stub(focusElement, 'focus', e => {
> ```

Why declare the global object? We've usually omitted it so for consistency's sake I'd still prefer omitting it in `<vaadin-input>` also

---

*[test/control-state.html, line 85 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-KcwAzmTCjpgd4AlwXxo:-KcwAzmUeTb-QXoJwOrf:b-cfv7o7) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/test/control-state.html#L85)):*
> ```HTML
> 
>         it('should blur the internal input on blur', function(done) {
>           window.sinon.stub(focusElement, 'focus', e => {
> ```

Move all of this inside a `beforeEach` part (inside 'focus'). Also the setTimeout should be moved there so that only thing that remains in the test case is:
```
  customElement._focus();
  customElement._blur();
  expect(customElement.focused).to.be.false;
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

platosha:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 5 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw2d65ioE_5zTaAil5:-KcwIC0PQ8f-ID5mYk40:b-yyi095) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L5)):*

Ouch, I didn't realize we're polluting the global namespace already. Our stuff should probably be under `vaadin` namespace?
</blockquote></details>

Right, `Vaadin.` namespace.

This was discussed here: https://www.flowdock.com/app/vaadin/elements-team/threads/e54rD71FuILZFg4cfayOLVYuaGn

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

Saulis:
Review status: 5 of 15 files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 49 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7jGmtqkZmcFQkgPq:-Kd0cX6f1WpEQf0h5PeW:bpu0nmp) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L49)):*

This relies on the implementer element to implement the `focus()` method. 

Maybe give non-implemented warning in console?

Alternatively, making `_focus()` public, as suggested below, will also resolve this.
</blockquote></details>

I'd say we just need to make sure the element is focusable so that calling the native `focus()` will fire `focus` event which will then be handled in `_focus()`.

---

*[vaadin-control-state-mixin.html, line 36 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd0c-QTIgKHCpaBRJUY:-Kd0c-QTIgKHCpaBRJUZ:b-sm73vo) ([raw file](https://github.com/vaadin/vaadin-input/blob/582ec6a1cea3fc4ced0a856706608a0d241f1ce4/vaadin-control-state-mixin.html#L36)):*
> ```HTML
>            * @private
>            */
>           tabindex: {
> ```

Like native inputs, `<vaadin-input>` should be focusable (tabindex = 0) by default. Default value can be set in `vaadin-input.html` or somewhere else as long as it is set.

---

*[vaadin-control-state-mixin.html, line 107 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd0a22tj_EoA11zf0nI:-Kd0a22tj_EoA11zf0nJ:bk41pa0) ([raw file](https://github.com/vaadin/vaadin-input/blob/582ec6a1cea3fc4ced0a856706608a0d241f1ce4/vaadin-control-state-mixin.html#L107)):*
> ```HTML
>         }
>         if (window.ShadyDOM) {
>           this._setTabindex(-1);
> ```

Why `-1` ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

manolo:
Review status: 5 of 15 files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 5 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw2d65ioE_5zTaAil5:-Kd0_5_X6yP87il7mNuU:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L5)):*

Right, `Vaadin.` namespace.

This was discussed here: https://www.flowdock.com/app/vaadin/elements-team/threads/e54rD71FuILZFg4cfayOLVYuaGn
</blockquote></details>

Done.

---

*[vaadin-control-state-mixin.html, line 14 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw-8a4NXsJsLTuJmM9:-Kd0_947CmcoMMAZLHW4:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L14)):*

Perhaps it is unnecessary to specify the `false` value.

If you do specify, though, then `disabled` should also have it for consistency.
</blockquote></details>

Done.

---

*[vaadin-control-state-mixin.html, line 36 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw1Hf0d-vOqUTGaHtm:-Kd0_AXpIFtt8iopcJtm:b-qg69yp) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L36)):*

We can’t use `attributeChangedCallback`/`observedAttributes` for that, right?

Maybe document that developers can use `tabindex` attribute instead of declaring as “Internal” and “@private” then?
</blockquote></details>

As discussed face to face the observedAttributes static getter is not working well in Polymer 2 when overridden. 

---

*[vaadin-control-state-mixin.html, line 56 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw4MS527DucwRDF16Z:-Kd0gAUZiRO-wSFtpsd4:bc1966s) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L56)):*

if the implementer element happens to define `_onElemFocus` method as well, will the implementer method used here then?

If so, we should maybe find a better way of protecting these event handler methods from accidental override.
</blockquote></details>

Let discuss this with the team whether makes sense to enclose mixins in scoped functions.
Right now inlining the callbacks since they are really simple. We can do protection in separated PR in the case.

---

*[vaadin-control-state-mixin.html, line 62 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw5UZmgVZmjQeoQ_pq:-Kd0pfQHOnMvXXKiw0u2:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L62)):*

If this is a mixin’s API, it should be documented.

Btw, maybe a getter method?
</blockquote></details>

Done.

---

*[vaadin-control-state-mixin.html, line 67 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw5CXcLFZCOfQM9gD4:-Kd1-YfHStNqs_oR0UpX:b-7ohoz9) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L67)):*

Why "Elem" instead of "Element"? Let's use "Element" everywhere for consistency.
</blockquote></details>

inlined, non needed anymore.

---

*[vaadin-control-state-mixin.html, line 49 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7jGmtqkZmcFQkgPq:-Kd177gjGzAR_HJuueyp:b-3fzy5x) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L49)):*

I'd say we just need to make sure the element is focusable so that calling the native `focus()` will fire `focus` event which will then be handled in `_focus()`.
</blockquote></details>

Making it public, I think it's the best 

---

*[vaadin-control-state-mixin.html, line 71 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7Txh5hR1AaJAlbVW:-Kd1-cCUJfPGvO0OfTJD:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L71)):*

Maybe expose `focus()` and `blur()` as public right away from here?
</blockquote></details>

Done.

---

*[vaadin-control-state-mixin.html, line 36 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd0c-QTIgKHCpaBRJUY:-Kd16Oi_vAvHHzB-Ye8R:b-pnoaha) ([raw file](https://github.com/vaadin/vaadin-input/blob/582ec6a1cea3fc4ced0a856706608a0d241f1ce4/vaadin-control-state-mixin.html#L36)):*

Like native inputs, `<vaadin-input>` should be focusable (tabindex = 0) by default. Default value can be set in `vaadin-input.html` or somewhere else as long as it is set.
</blockquote></details>

Good point, it should only happen in shadow, and cannot be done by reflect to attribute property, so accomplishing it in the ready callback. 

Done.

---

*[vaadin-control-state-mixin.html, line 107 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd0a22tj_EoA11zf0nI:-Kd1-dVXmTFFBkOE-qol:b-yw8ilj) ([raw file](https://github.com/vaadin/vaadin-input/blob/582ec6a1cea3fc4ced0a856706608a0d241f1ce4/vaadin-control-state-mixin.html#L107)):*

Why `-1` ?
</blockquote></details>

Otherwise we don't have the native `.focus()`, but since now we provide the focus method in the mixin, we don't need it anymore here.
Removed.

---

*[test/control-state.html, line 31 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw6mNxacgZ6Qp6W8Gf:-Kd11KFnbPEeXZjSiA_C:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/test/control-state.html#L31)):*

Let us combine this test with the other shady-DOM soped tests below in a single `if ... else ... `. This way would be easier to see what is different.
</blockquote></details>

Done.

---

*[test/control-state.html, line 73 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-KcwBOzaJ26Ih4K9OUrP:-Kd14RSO__5Kom1CL4x5:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/test/control-state.html#L73)):*

Why declare the global object? We've usually omitted it so for consistency's sake I'd still prefer omitting it in `<vaadin-input>` also
</blockquote></details>

Done.

---

*[test/control-state.html, line 85 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-KcwAzmTCjpgd4AlwXxo:-Kd14QpaeU_Q0bHFTiDy:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/test/control-state.html#L85)):*

Move all of this inside a `beforeEach` part (inside 'focus'). Also the setTimeout should be moved there so that only thing that remains in the test case is:
```
  customElement._focus();
  customElement._blur();
  expect(customElement.focused).to.be.false;
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

platosha:
Reviewed 8 of 10 files at r3, 6 of 7 files at r4, 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 16 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 14 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw-8a4NXsJsLTuJmM9:-Kd1Jh4lhCSIw2_77DLL:bfhqml6) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L14)):*

Done.
</blockquote></details>

Not done.

---

*[vaadin-control-state-mixin.html, line 36 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw1Hf0d-vOqUTGaHtm:-Kd0jtaGOV5ymOUYIQg_:b-th9wiu) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L36)):*

We can’t use `attributeChangedCallback`/`observedAttributes` for that, right?

Maybe document that developers can use `tabindex` attribute instead of declaring as “Internal” and “@private” then?
</blockquote></details>

LGTM as discussed and prototyped, `observedAttributes` is not a nice option to use.

---

*[vaadin-control-state-mixin.html, line 56 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw4MS527DucwRDF16Z:-Kd0kId58Nw9VQ6czOYV:b6dn75v) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L56)):*

if the implementer element happens to define `_onElemFocus` method as well, will the implementer method used here then?

If so, we should maybe find a better way of protecting these event handler methods from accidental override.
</blockquote></details>

LGTM

This probably should be considered for all the 2.0 mixins, outside the scope of this PR. Agreed to discuss separately.

---

*[vaadin-control-state-mixin.html, line 62 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw5UZmgVZmjQeoQ_pq:-Kd1Jxh1lxIrRD6hO5uo:bfhqml6) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L62)):*

Done.
</blockquote></details>

Not done.

---

*[vaadin-control-state-mixin.html, line 49 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7jGmtqkZmcFQkgPq:-Kd0kACyctkbZh7T298W:bb6ywmx) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L49)):*

Making it public, I think it's the best 
</blockquote></details>

LGTM

---

*[vaadin-control-state-mixin.html, line 71 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7Txh5hR1AaJAlbVW:-Kd0krgutjO4U4wscQ6L:b-4bsqvb) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L71)):*

Maybe expose `focus()` and `blur()` as public right away from here?
</blockquote></details>

LGTM

OK for me to have them underscored now.

---

*[vaadin-control-state-mixin.html, line 6 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd1JWezntJ-OODbx0fw:-Kd1JWf-iYUDyvSNcHKo:b8leyg5) ([raw file](https://github.com/vaadin/vaadin-input/blob/50e4b7ef2761363cfaf0dba6bb34599dd5a94134/vaadin-control-state-mixin.html#L6)):*
> ```HTML
>   // Polymer.IronControlState is not a proper 2.0 class, also, its tabindex
>   // implementation fails in the shadow dom, so we have this for vaadin elements.
>   Vaadin.ControlStateMixin = subclass => class ControlStateMixin extends subclass {
> ```

Maybe `VaadinControlStateMixin`.

---

*[vaadin-form-element-mixin.html, line 5 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd1E6LzGUbjwh_TJDdi:-Kd1E6LzGUbjwh_TJDdj:b-kjcaxx) ([raw file](https://github.com/vaadin/vaadin-input/blob/eb2630156dee5ff709b21f3a6d6c042453ab23dc/vaadin-form-element-mixin.html#L5)):*
> ```HTML
> 
>   // IronFormElementBehavior is not a proper 2.0 class.
>   Vaadin.FormElementMixin = subclass => class FormElementMixin extends subclass {
> ```

It is slightly better to have Vaadin in the mixin name as well:
`Vaadin.FormElementMixin = subclass => class VaadinFormElementMixin ...`

Reasoning: you would see `VaadinFormElementMixin` in developer tools debugger / console. Seeing just `FormElementMixin`, otherwise, would be less relevant and helpful.

---

*[vaadin-input.html, line 76 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd1EosTG3uULtcizCBW:-Kd1EosTG3uULtcizCBX:b-mplvvn) ([raw file](https://github.com/vaadin/vaadin-input/blob/eb2630156dee5ff709b21f3a6d6c042453ab23dc/vaadin-input.html#L76)):*
> ```HTML
>     window.Vaadin = window.Vaadin || {};
> 
>     Vaadin.InputElement = class extends Vaadin.ControlStateMixin(Vaadin.FormElementMixin(Polymer.Element)) {
> ```

`... = class VaadinInputElement ...` maybe? The same reason: it is nice to see in developer tools.

---

*[test/control-state.html, line 40 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd1KE4pROHU181_dc54:-Kd1KE4pROHU181_dc55:bopzync) ([raw file](https://github.com/vaadin/vaadin-input/blob/50e4b7ef2761363cfaf0dba6bb34599dd5a94134/test/control-state.html#L40)):*
> ```HTML
> 
>       describe('tabindex', function() {
>         if (window.ShadyDOM) {
> ```

Maybe have a nested `describe` blocks instead of having duplicate prefix `'in shady dom...'` in every test title?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

Saulis:
shift-tabbing requires 2x tabbing currently. 

---

Review status: all files reviewed at latest revision, 14 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 54 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd1Q4sGhlg7xqzTveKi:-Kd1Q4sHsYdMuKk7O7aU:b-chxjx1) ([raw file](https://github.com/vaadin/vaadin-input/blob/50e4b7ef2761363cfaf0dba6bb34599dd5a94134/vaadin-control-state-mixin.html#L54)):*
> ```HTML
> 
>       // In shadow dom, the element should be always focusable
>       if (!window.ShadyDOM && this.getAttribute('tabindex') == null) {
> ```

Well.. also in Shady vaadin-input needs to be focusable by default.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-:-Kd1Ppn5UUurtZ3aO4e-:b1mbd1m)*

Saulis:
Review status: all files reviewed at latest revision, 14 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 49 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7jGmtqkZmcFQkgPq:-Kd1NPlEfpZWQGoCxqEo:bi3gj6v) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L49)):*

LGTM
</blockquote></details>

I feel mixing native focus and overriding `focus()` is a bit dangerous – I think we should either override the `focus()` OR listen for native `focus` events on the host element.

Shift-tabbing from native input reliably out from host element requires us either to focus on the host element during the shift+tab key event OR making the host element non-focusable by removing the `tabindex`. Currently, the shift-tabbing doesn't work properly in Shadow because the host element is kept focusable, but we can't programmatically move focus to it because the native `focus()` has been overridden.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

Saulis:
Review status: all files reviewed at latest revision, 14 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 54 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd1Q4sGhlg7xqzTveKi:-Kd4bdfH0y1UiMApTl78:b4fngb8) ([raw file](https://github.com/vaadin/vaadin-input/blob/50e4b7ef2761363cfaf0dba6bb34599dd5a94134/vaadin-control-state-mixin.html#L54)):*

Well.. also in Shady vaadin-input needs to be focusable by default.
</blockquote></details>

Just to add, if we want `<vaadin-input>` to behave like native input, the host element needs to be focusable always expect when `disabled` to fire `focus` events.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

manolo:
Review status: 14 of 18 files reviewed at latest revision, 14 unresolved discussions.

---

*[vaadin-control-state-mixin.html, line 14 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw-8a4NXsJsLTuJmM9:-Kd2kHoIGRHEafWg-3b7:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L14)):*

Not done.
</blockquote></details>

Done.

---

*[vaadin-control-state-mixin.html, line 62 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw5UZmgVZmjQeoQ_pq:-Kd2lTIG5RGqrrX7JStu:b-ffb4i0) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L62)):*

Not done.
</blockquote></details>

I misunderstood, thought you wanted a public `getMethod` not a public property with a getter. Done

---

*[vaadin-control-state-mixin.html, line 49 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7jGmtqkZmcFQkgPq:-Kd0_u3iKZFTDhy1vDRD:b-3nvq7s) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L49)):*

I feel mixing native focus and overriding `focus()` is a bit dangerous – I think we should either override the `focus()` OR listen for native `focus` events on the host element.

Shift-tabbing from native input reliably out from host element requires us either to focus on the host element during the shift+tab key event OR making the host element non-focusable by removing the `tabindex`. Currently, the shift-tabbing doesn't work properly in Shadow because the host element is kept focusable, but we can't programmatically move focus to it because the native `focus()` has been overridden.
</blockquote></details>

@saulis You are right, we need to use the native `focus()` always. Rolled back.
Note that in the other hand, we don't need the native for blur, so keep overriding so as we dont need to listen for blur events.

---

*[vaadin-control-state-mixin.html, line 107 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd0a22tj_EoA11zf0nI:-Kd2lpXscrX0pQNZqyZH:b-k9n17y) ([raw file](https://github.com/vaadin/vaadin-input/blob/582ec6a1cea3fc4ced0a856706608a0d241f1ce4/vaadin-control-state-mixin.html#L107)):*

Otherwise we don't have the native `.focus()`, but since now we provide the focus method in the mixin, we don't need it anymore here.
Removed.
</blockquote></details>

We need -1 again, because we have native `vaadininput.focus()`  which means that the user can call the `focus()` programatically, so a tabindex is needed.

---

*[vaadin-control-state-mixin.html, line 6 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd1JWezntJ-OODbx0fw:-Kd2k5VxjlQ3hdTackVd:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/50e4b7ef2761363cfaf0dba6bb34599dd5a94134/vaadin-control-state-mixin.html#L6)):*

Maybe `VaadinControlStateMixin`.
</blockquote></details>

Done.

---

*[vaadin-control-state-mixin.html, line 54 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd1Q4sGhlg7xqzTveKi:-Kd2l8W1eUDXNK-ZACNs:btf2dvn) ([raw file](https://github.com/vaadin/vaadin-input/blob/50e4b7ef2761363cfaf0dba6bb34599dd5a94134/vaadin-control-state-mixin.html#L54)):*

Just to add, if we want `<vaadin-input>` to behave like native input, the host element needs to be focusable always expect when `disabled` to fire `focus` events.
</blockquote></details>

Since I rolled back to native focus, we need. Done.

---

*[vaadin-form-element-mixin.html, line 5 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd1E6LzGUbjwh_TJDdi:-Kd2m8WbO8ns-j9arVxD:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/eb2630156dee5ff709b21f3a6d6c042453ab23dc/vaadin-form-element-mixin.html#L5)):*

It is slightly better to have Vaadin in the mixin name as well:
`Vaadin.FormElementMixin = subclass => class VaadinFormElementMixin ...`

Reasoning: you would see `VaadinFormElementMixin` in developer tools debugger / console. Seeing just `FormElementMixin`, otherwise, would be less relevant and helpful.
</blockquote></details>

Done.

---

*[vaadin-input.html, line 76 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd1EosTG3uULtcizCBW:-Kd2mA8ptZ2yJSxcETax:b-ahuzui) ([raw file](https://github.com/vaadin/vaadin-input/blob/eb2630156dee5ff709b21f3a6d6c042453ab23dc/vaadin-input.html#L76)):*

`... = class VaadinInputElement ...` maybe? The same reason: it is nice to see in developer tools.
</blockquote></details>

In the element we need the namespace in the window, otherwise Hydrolysis does not generate any apidoc.  Rolled back until they fix the issue.

---

*[test/control-state.html, line 40 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd1KE4pROHU181_dc54:-Kd4gZM07PUuqcIFh-Gh:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/50e4b7ef2761363cfaf0dba6bb34599dd5a94134/test/control-state.html#L40)):*

Maybe have a nested `describe` blocks instead of having duplicate prefix `'in shady dom...'` in every test title?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

Saulis:
Review status: 14 of 18 files reviewed at latest revision, 14 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 49 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7jGmtqkZmcFQkgPq:-Kd55D6qsMlZSottvO7_:buy8j19) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L49)):*

@saulis You are right, we need to use the native `focus()` always. Rolled back.
Note that in the other hand, we don't need the native for blur, so keep overriding so as we dont need to listen for blur events.
</blockquote></details>

Blur needs to also be native, otherwise, we don't get the event in some cases.

Try with this snippet on the first demo example:
```html
<vaadin-input value="foo"></vaadin-input>
<script>
var input = document.querySelector('vaadin-input');
  input.addEventListener('focus', (e) => {
    console.log('focus', e.target);
  });

  input.addEventListener('blur', (e) => {
    console.log('blur', e.target);
  });
</script>
```

Each focus and blur action (whether it be with tabbing or clicking) should fire exactly one event. This needs to work the same way across different browsers and shady + shadow.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

platosha:
Reviewed 4 of 4 files at r6.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 62 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw5UZmgVZmjQeoQ_pq:-Kd5l8XE0uxo2Ee_1Yok:b-hyy335) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L62)):*

I misunderstood, thought you wanted a public `getMethod` not a public property with a getter. Done
</blockquote></details>

Not quite done, sorry to confuse. I was asking to have an APIdoc here.

It is required to implement this thing in the element, therefore this is an API of the mixin, that element developer consumes, right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

Saulis:
Haven't tried on other browsers than Chrome and Firefox, but this stripped down version seems to work for me:

```html
<script>
  window.Vaadin = window.Vaadin || {};

  // Polymer.IronControlState is not a proper 2.0 class, also, its tabindex
  // implementation fails in the shadow dom, so we have this for vaadin elements.
  Vaadin.ControlStateMixin = subclass => class VaadinControlStateMixin extends subclass {
    static get config() {
      return {
        properties: {
          /**
           * If true, the element currently has focus.
           */
          focused: {
            type: Boolean,
            notify: true,
            readOnly: true,
            observer: 'focusedChanged',
            reflectToAttribute: true
          },

          /**
           * If true, the user cannot interact with this element.
           */
          disabled: {
            type: Boolean,
            observer: '_disabledChanged',
            reflectToAttribute: true
          },

          /**
           * Internal property needed to listen to `tabindex` attribute changes.
           *
           * For changing the tabindex of this component use the native `tabIndex` propety.
           * @private
           */
          tabindex: {
            type: Number,
            value: 0
          }
        }
      };
    }

    ready() {
      super.ready();

      // Make FF take the focus (#9)
      if (this.autofocus) {
        this.focus();
      }

      this.addEventListener('focus', e => this._focus());
      this.addEventListener('blur', e => this._blur());

      this.addEventListener('keydown', e => {
        if (e.shiftKey && e.keyCode === 9) {
          this.focus();
        }
      });
    }

    get focusElement() {
      window.console.warn(`Please implement the 'focusElement' property in <${this.localName}>`);
      return this;
    }

    _focus() {
      this._setFocused(true);
      this.focusElement.focus();
    }

    _blur() {
      this._setFocused(false);
      this.focusElement.blur();
    }

    _disabledChanged(disabled) {
      this.focusElement.disabled = disabled;
      if (disabled) {
        this.blur();
        this.removeAttribute('tabindex');
      } else {
        this.setAttribute('tabindex', this.tabindex);
      }
    }

    focusedChanged() {
      // Needed for Edge, until they have support for CSS Custom Properties
      // (already shipping in Edge Preview)
      this.updateStyles();
    }
  };
</script>
```

---

Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-:-Kd65PxDHTzpyPzRniEb:bhsx4cx)*

manolo:
There are some cases not covered by this implementation. And other issues that we can discuss face-to-face.
I have updated test cases with almost features so as you can try those with this implementation.

---

Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

*[vaadin-control-state-mixin.html, line 62 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw5UZmgVZmjQeoQ_pq:-Kd8Gg-pU1WKt7Se-r1m:b-896fix) ([raw file](https://github.com/vaadin/vaadin-input/blob/1151782704fbb5785c2cddd70ebc7040519fb276/vaadin-control-state-mixin.html#L62)):*

Not quite done, sorry to confuse. I was asking to have an APIdoc here.

It is required to implement this thing in the element, therefore this is an API of the mixin, that element developer consumes, right?
</blockquote></details>

Done.

---

*[vaadin-control-state-mixin.html, line 49 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7jGmtqkZmcFQkgPq:-Kd8ERqbsw2FuxPkETVf:b-6ehavk) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L49)):*

Blur needs to also be native, otherwise, we don't get the event in some cases.

Try with this snippet on the first demo example:
```html
<vaadin-input value="foo"></vaadin-input>
<script>
var input = document.querySelector('vaadin-input');
  input.addEventListener('focus', (e) => {
    console.log('focus', e.target);
  });

  input.addEventListener('blur', (e) => {
    console.log('blur', e.target);
  });
</script>
```

Each focus and blur action (whether it be with tabbing or clicking) should fire exactly one event. This needs to work the same way across different browsers and shady + shadow.
</blockquote></details>

Blur cannot be native, otherwise user cannot call `vaadin_input.blur()` when focus is in `focusElement`, since blur only works on the element which has the focus. Try it in FF or Edge. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-:-Kd8G6ha8ZfbzMTTQUiz:b5mbbek)*

manolo:
Review status: 8 of 25 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[vaadin-control-state-mixin.html, line 49 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kcw7jGmtqkZmcFQkgPq:-Kd8H_AqKge-cLg2prO9:b-sx0yqf) ([raw file](https://github.com/vaadin/vaadin-input/blob/d74d283b8326bb5d72c3906f7ba0e3a791392242/vaadin-control-state-mixin.html#L49)):*

Blur cannot be native, otherwise user cannot call `vaadin_input.blur()` when focus is in `focusElement`, since blur only works on the element which has the focus. Try it in FF or Edge. 
</blockquote></details>

The  way to cover this case is dispatching the event manually, but I don't think we should do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

manolo:
Review status: 8 of 23 files reviewed at latest revision, 9 unresolved discussions.

---

*[vaadin-control-state-mixin.html, line 107 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-input/17#-Kd0a22tj_EoA11zf0nI:-KdAxmOs23RN0uFj_abw:bb6un1j) ([raw file](https://github.com/vaadin/vaadin-input/blob/582ec6a1cea3fc4ced0a856706608a0d241f1ce4/vaadin-control-state-mixin.html#L107)):*

We need -1 again, because we have native `vaadininput.focus()`  which means that the user can call the `focus()` programatically, so a tabindex is needed.
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-input/17)*

PR #51, opened 2017-04-01
Prefix and suffix feature

### LINKED ISSUE #25: Prefix and Suffix
Use Cases to cover:
- allow adding icons/buttons inside the focusable area (think of combo-box)
- consider disabling "floating label" when prefix is defined

tomivirkki:
Closes #25

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/51)


Saulis:
:lgtm:

---

Reviewed 16 of 16 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/51#-:-Kgn_SnKXwcvXe6ZdrVt:bnfp4nl)*

limonte:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/51#-:-KgncgVQhz9kIH7oMYl3:bnfp4nl)*

manolo:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/51#-:-KgnfkJ--GTcklMkUI6K:bnfp4nl)*

jouni:
Nice, you made it work with floating labels as well 👍 

But why the use of id’s in the template elements? We aren’t even using the Polymer convenience accessor for those (`this.$`), and it just pollutes the DOM with multiple same id’s in polyfilled browsers. I think it would have been future-proof to use the `part="suffix"` syntax already, as we need to change to that anyway before releasing the first version.

jouni:
Using `part` of course means we need to decide which parts to actually expose, so it was probably quicker to implement this way.

tomivirkki:
We use id’s pretty much overall in templates to mark unique elements. There were some in this template also before this ticket and those of the new ones are mostly derived from `<paper-input-container>`. Didn’t consider switching to using the part markers as part of this specific ticket. But sure it should be done before 1.0.

jouni:
Sure, no problem :) I’ve just always wondered this pattern, as I feel the id’s are so “heavy-weight” because of their CSS specificity and the fact that you should only have unique id’s in the DOM (we might be breaking someone’s app if they happen use the same exact id for their app specific element, right?).

Issue #48, opened 2017-03-31
Accessibility support


Saulis:
For reference: https://www.w3.org/TR/wai-aria-1.1/#textbox

Proposal: should support announcing label by the screenreader, also should announce states like disabled, readonly, invalid and required (and error message).

Also the value should be announced :doh:. And should announce that the user is in a textbox.

All the announcements should work out of the box without requiring the users to add any aria attributes themselves. `aria-label` can of course be used to override the default announcement.

Should be tested on VoiceOver (OSX + iOS), ChromeVox, JAWS and NVDA on Windows 10, Android TalkBack

Saulis:
Notes from grooming: 
- try to limit the scope of this ticket so that no changes are required for the existing error message functionality.

PR #56, opened 2017-04-06
Adding errorMessage

### LINKED ISSUE #26: Error message
Specs by @jouni

It should be possible to provide an error message for the text-field, which gets announced by the screen reader, for example in the case of a validation error.

The error message should only be displayed when the text field is invalid.

**API proposal**:
```html
<vaadin-text-field error-message="This field is required" label="Name" required></vaadin-text-field>
```

**Implementation details**

Expose the error message as a new themable part, e.g. 
```html
# shadow root
  ...
  <div class="error" part="error-message">[[errorMessage]]</div>
  ...
```

The error message should be displayed under the field by default. No extra styling is needed (colors or anything else).

manolo:
Fixes #26

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/56)


limonte:
In ideal world, you should create a PR after green branch build :) Take a look on tests.

manolo:
In ideal world, creating a PR mustn't move a task to review column

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-:-Kh22KLH3t3i47pMZm4n:b4y93e9)*

tomivirkki:
The implementation could be more declarative. Here’s what I suggest: https://github.com/vaadin/vaadin-text-field/tree/feature/error-message-suggestions

---

Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-text-field.html, line 98 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2BZhUnXt4l7tN4MPc:-Kh2BZhVTSVmFZELfSrQ:b-dwfn1m) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b33ba1ede3d2484577767d5a6c8ed25c4e761b2b/vaadin-text-field.html#L98)):*
> ```HTML
> 
>       .error {
>         visibility: hidden;
> ```

Using `visibility` to hide the error makes the textfield box stretch depending on the length of the error message even if it's invisible.

---

*[demo/index.html, line 122 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2Ygf1dLKxSmRumoYL:-Kh2Ygf2lfuX-9cCd1Xy:b7k9dtr) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/demo/index.html#L122)):*
> ```HTML
>     </demo-snippet>
> 
>     <div class="form-item">
> ```

Should this be inside a demo snippet?

---

*[test/accessibility.html, line 26 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2YlxAnWvuMa0TEix0:-Kh2YlxAnWvuMa0TEix1:b6wqqct) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/test/accessibility.html#L26)):*
> ```HTML
>       beforeEach(function(done) {
>         tf = fixture('error-fixture');
>         setTimeout(done);
> ```

Don't use setTimeout

---

*[test/accessibility.html, line 46 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2YnrYKumpF0Yc6oxO:-Kh2YnrYKumpF0Yc6oxP:b6wqqct) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/test/accessibility.html#L46)):*
> ```HTML
>         tf.validate();
> 
>         setTimeout(() => {
> ```

Don't use setTimeout

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-:-Kh2ZAPSuVRLoDsH5OkI:bni9dr)*

platosha:
Reviewed 1 of 4 files at r1, 1 of 2 files at r3.
Review status: 2 of 4 files reviewed at latest revision, 9 unresolved discussions.

---

*[vaadin-text-field.html, line 324 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2ZeGo_rEuWWumsvQB:-Kh2ZeGo_rEuWWumsvQC:b2o7zpm) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/vaadin-text-field.html#L324)):*
> ```HTML
>           setTimeout(() => {
>             // Timeout so as DOM has been updated
>             this._error = Polymer.dom(this.root).querySelector('#error');
> ```

We can always find the error element by classname, maybe no need to memoize.

---

*[vaadin-text-field.html, line 328 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2_INySh-D818uaN57:-Kh2_INySh-D818uaN58:b26p1ze) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/vaadin-text-field.html#L328)):*
> ```HTML
>             // In shady we need unique ids in the document
>             if (window.ShadyDOM) {
>               this._error.id += (VaadinTextFieldElement.uniqueId = 1 + VaadinTextFieldElement.uniqueId || 0);
> ```

`uniqueId` should be private.

Changing the ids on the fly is not expected.

Maybe we should bind the `aria-describedby` attribute and error id in the template,  that would be nicer to read.

---

*[demo/index.html, line 118 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2XcUZScQNLCh24EFj:-Kh2XcUZScQNLCh24EFk:b-fgqz2e) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/demo/index.html#L118)):*
> ```HTML
>     <demo-snippet>
>       <template>
>         <p><vaadin-text-field required errorMessage="You must write something in this input"></vaadin-text-field> Can type numbers only</p>
> ```

camelCase `errorMessage` attribute, should be kebab-case.

---

*[demo/index.html, line 125 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2aNZzryaFDDoRi8kF:-Kh2aNZzryaFDDoRi8kG:bolzrtz) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/demo/index.html#L125)):*
> ```HTML
>       <vaadin-text-field label="User Name"
>         minlength="4" maxlength="8" required pattern="^\w+$"
>         error-message="from 4 to 8 characters without spaces">
> ```

What’s the value of this demo, compared to the previous one above?

---

*[demo/index.html, line 129 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2YTUQfc5HIN7_1VXo:-Kh2YTUQfc5HIN7_1VXp:b-zfzbsu) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/demo/index.html#L129)):*
> ```HTML
>       </vaadin-text-field>
>     </div>
> ```

Can we have a demo with external validation that assigns the error message when invalid?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

manolo:
@tomivirkki good point, using declarative which makes code simpler

---

Review status: 0 of 4 files reviewed at latest revision, 9 unresolved discussions.

---

*[vaadin-text-field.html, line 98 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2BZhUnXt4l7tN4MPc:-Kh2glFjuazu2LFG6slx:bnxfkbq) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b33ba1ede3d2484577767d5a6c8ed25c4e761b2b/vaadin-text-field.html#L98)):*

Using `visibility` to hide the error makes the textfield box stretch depending on the length of the error message even if it's invisible.
</blockquote></details>

That is the approach used in paper-input as well

---

*[vaadin-text-field.html, line 324 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2ZeGo_rEuWWumsvQB:-Kh2gQ9YFjr3BNMci8ca:bv4ifh9) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/vaadin-text-field.html#L324)):*

We can always find the error element by classname, maybe no need to memoize.
</blockquote></details>

Not needed to transverse the dom with the new approach

---

*[vaadin-text-field.html, line 328 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2_INySh-D818uaN57:-Kh2gD_qGobHTmp7TqnO:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/vaadin-text-field.html#L328)):*

`uniqueId` should be private.

Changing the ids on the fly is not expected.

Maybe we should bind the `aria-describedby` attribute and error id in the template,  that would be nicer to read.
</blockquote></details>

Done.

---

*[demo/index.html, line 118 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2XcUZScQNLCh24EFj:-Kh2hBRdgE6UiECxocPH:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/demo/index.html#L118)):*

camelCase `errorMessage` attribute, should be kebab-case.
</blockquote></details>

Done.

---

*[demo/index.html, line 122 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2Ygf1dLKxSmRumoYL:-Kh2hCuei1S_sh5Frnq3:beebutg) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/demo/index.html#L122)):*

Should this be inside a demo snippet?
</blockquote></details>

This demo shouldn't be here, just committed by mistake

---

*[demo/index.html, line 125 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2aNZzryaFDDoRi8kF:-Kh2hU5wJUu-09jUAqv3:blegzx1) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/demo/index.html#L125)):*

What’s the value of this demo, compared to the previous one above?
</blockquote></details>

No value, removed

---

*[demo/index.html, line 129 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2YTUQfc5HIN7_1VXo:-Kh2hVKLcqutOppN3P9e:b-690lfb) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/demo/index.html#L129)):*

Can we have a demo with external validation that assigns the error message when invalid?
</blockquote></details>

Then it should go to the custom validation demo. But I think with this demo is enough to the user know that there is a property for errors.

---

*[test/accessibility.html, line 26 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2YlxAnWvuMa0TEix0:-Kh2jpYzJPbvhN6cQjdO:b-7kflox) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/test/accessibility.html#L26)):*

Don't use setTimeout
</blockquote></details>

Async is necessary for these tests, based on P2.0 doc just animation-frame or timeout should be enough.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-:-Kh2gbWPAQ5WaGAuu7cV:bdh2336)*

platosha:
Reviewed 1 of 4 files at r1, 2 of 2 files at r4, 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-Kh2mfZCczDp17YhAhlj:brcjwce) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*
> ```HTML
>         <slot name="suffix"></slot>
> 
>         <template id="errortpl" is="dom-if" if="[[errorMessage]]">
> ```

`<tempalte is="dom-if">` is a legacy syntax. We should probably avoid `is` in P2 elements.

---

*[vaadin-text-field.html, line 318 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2kqkKu2BZT3KLjYf-:-Kh2kqkKu2BZT3KLjYf0:b23ttlh) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L318)):*
> ```HTML
>         // ShadyDOM needs unique ids in the document
>         var uniqueId = window.ShadyDOM ? VaadinTextFieldElement._uniqueId = 1 + VaadinTextFieldElement._uniqueId || 0 : '';
>         this._errorId = 'text-field-error' + uniqueId;
> ```

BTW I would prefix with `vaadin-text-field-error`. Not a blocker.

---

*[demo/index.html, line 129 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2YTUQfc5HIN7_1VXo:-Kh2lDncksBnVvz0p9QN:b40n129) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/demo/index.html#L129)):*

Then it should go to the custom validation demo. But I think with this demo is enough to the user know that there is a property for errors.
</blockquote></details>

OK in to have in another demo. I wonder how accessibility works in that use case. 

Would like us to make sure the error message is spoken when set, i. e., that `aria-live` has meaningful effect. A dynamic error message demo would help us to check that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

Saulis:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-Kh3TprEXZTgqCQq-nId:b9eeutp) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*

`<tempalte is="dom-if">` is a legacy syntax. We should probably avoid `is` in P2 elements.
</blockquote></details>

Also, the error message is currently announced even if it's hidden.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

tomivirkki:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-text-field.html, line 98 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2BZhUnXt4l7tN4MPc:-Kh5UoRlg-XVdRQtkQjL:b-5xtq6b) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b33ba1ede3d2484577767d5a6c8ed25c4e761b2b/vaadin-text-field.html#L98)):*

That is the approach used in paper-input as well
</blockquote></details>

Sure, but in their implementation `<paper-input-error>` also has `position: absolute` so it doesn't affect the layout.

We should also probably move the error div inside the `.container` so it's below the input.

---

*[vaadin-text-field.html, line 73 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5bQTfrkr3aa19_-xW:-Kh5bQTfrkr3aa19_-xX:b-m6bznf) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L73)):*
> ```HTML
>       }
> 
>       .error {
> ```

The error message is something that developers should be able to style so switch from `class="error"`to `part="error"`

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh5gxC5DFHL-79o-2yE:b-iojbig) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*
> ```HTML
>       }
> 
>       :host([invalid]) .error {
> ```

These rules can be combined to
```
:host(:not([invalid])) .error {
  visibility: hidden;
}
```
...or removed completely if you use `hidden=[[_hideError(errorMessage, invalid)]]` in the template

---

*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-Kh5UtoSe7V-Y4iIdHHN:b-t2bd5u) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*

Also, the error message is currently announced even if it's hidden.
</blockquote></details>

Can be shortened to `hidden="[[!errorMessage]]"`.

---

*[test/accessibility.html, line 26 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2YlxAnWvuMa0TEix0:-Kh5gOU7WTegU07zlUN5:b8os7ib) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/test/accessibility.html#L26)):*

Async is necessary for these tests, based on P2.0 doc just animation-frame or timeout should be enough.
</blockquote></details>

Not necessary, just add `Polymer.dom.flush();` here before the "querySelector" line and in the other test before the "expect" line

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
*[vaadin-text-field.html, line 73 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5bQTfrkr3aa19_-xW:-Kh5xQ1eCCIUf08Kgi5Q:b-uscj9r) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L73)):*

The error message is something that developers should be able to style so switch from `class="error"`to `part="error"`
</blockquote></details>

`part="error-message"` would be more appropriate.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

manolo:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-text-field.html, line 98 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2BZhUnXt4l7tN4MPc:-Kh6PC9sTYSvzTp0r9ek:b-3gf9n7) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b33ba1ede3d2484577767d5a6c8ed25c4e761b2b/vaadin-text-field.html#L98)):*

Sure, but in their implementation `<paper-input-error>` also has `position: absolute` so it doesn't affect the layout.

We should also probably move the error div inside the `.container` so it's below the input.
</blockquote></details>

Moved to the container, for this case it seems better to have visibility=none since there are no layout modifications when showing/hiding the message. Let's theme designers to modify defaults if they need.

---

*[vaadin-text-field.html, line 73 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5bQTfrkr3aa19_-xW:-Kh60kFVPOqltp15-ELn:bn34y1o) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L73)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

`part="error-message"` would be more appropriate.
</blockquote></details>

Done, I didn't know we were already using this selector in our components.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh6OesSpUDIRtmltt69:b-u964z9) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

These rules can be combined to
```
:host(:not([invalid])) .error {
  visibility: hidden;
}
```
...or removed completely if you use `hidden=[[_hideError(errorMessage, invalid)]]` in the template
</blockquote></details>

This is the same case, whether we prefer that the layout dimension changes by default or not. Leaving it in the css that also should give designers the ability to play with the element.

---

*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-Kh5xv095a3Tq5nApqaa:bocpw3s) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*

Can be shortened to `hidden="[[!errorMessage]]"`.
</blockquote></details>

apparently dom-if does not render the content if error message is not set which is an optimisation in most use cases where the user does not uses a custom error message. Also it's how paper-input does.

---

*[vaadin-text-field.html, line 318 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2kqkKu2BZT3KLjYf-:-Kh6RGciMqDHBHZ_t_OY:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L318)):*

BTW I would prefix with `vaadin-text-field-error`. Not a blocker.
</blockquote></details>

Done.

---

*[demo/index.html, line 129 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2YTUQfc5HIN7_1VXo:-Kh6VnB0_YqDSgCqLWRI:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/demo/index.html#L129)):*

OK in to have in another demo. I wonder how accessibility works in that use case. 

Would like us to make sure the error message is spoken when set, i. e., that `aria-live` has meaningful effect. A dynamic error message demo would help us to check that.
</blockquote></details>

Done.

---

*[test/accessibility.html, line 26 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2YlxAnWvuMa0TEix0:-Kh6Ro9TYSBQg2Bd64OT:b-u55id9) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/test/accessibility.html#L26)):*

Not necessary, just add `Polymer.dom.flush();` here before the "querySelector" line and in the other test before the "expect" line
</blockquote></details>

Sorry, didn't understood what you meant. Done

---

*[test/accessibility.html, line 46 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2YnrYKumpF0Yc6oxO:-Kh6Rsrgnx86gxHNFMAm:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a3ad1fa873622e08babe7fd7c928b21bdd7a439f/test/accessibility.html#L46)):*

Don't use setTimeout
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

tomivirkki:
Review status: 1 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[vaadin-text-field.html, line 98 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2BZhUnXt4l7tN4MPc:-Kh6Zk34VIa4JqGmp1Mj:blphx9l) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b33ba1ede3d2484577767d5a6c8ed25c4e761b2b/vaadin-text-field.html#L98)):*

Moved to the container, for this case it seems better to have visibility=none since there are no layout modifications when showing/hiding the message. Let's theme designers to modify defaults if they need.
</blockquote></details>

It still has the original problem: the error message length affects the element width (even when hidden). Fixing this is not the responsibility of themes.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh6_G8P9ItTINIJYmDc:bg2m4dx) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*
> Leaving it in the css that also should give designers the ability to play with the element

This doesn't affect the designers in any way. As long as the element has "part" defined, designers can target it.

---

*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-Kh6a2FrhIMTEj0FMNim:b-xs06br) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*

apparently dom-if does not render the content if error message is not set which is an optimisation in most use cases where the user does not uses a custom error message. Also it's how paper-input does.
</blockquote></details>

`dom-if` is actually heavier than what I suggested because it adds 2 (non-displayed) elements in the dom (`<dom-if>` and `<template>`) and adds another one when the condition value changes. `hidden="[[!errorMessage]]"` only adds 1.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

manolo:
Review status: 1 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[vaadin-text-field.html, line 98 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2BZhUnXt4l7tN4MPc:-Kh6d5-3BaIZfumfMPHj:b8tzft9) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b33ba1ede3d2484577767d5a6c8ed25c4e761b2b/vaadin-text-field.html#L98)):*

It still has the original problem: the error message length affects the element width (even when hidden). Fixing this is not the responsibility of themes.
</blockquote></details>

Still not convinced, why is better that is good that the apparition of the error message should affect the layout ? but the hidden does not reserve space. Both cases can be addressed by the developer since we provide the selector, so is a question of taste I guess.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh6dxa0GDNx9-Bet87L:br61j6w) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

> Leaving it in the css that also should give designers the ability to play with the element

This doesn't affect the designers in any way. As long as the element has "part" defined, designers can target it.
</blockquote></details>

I don't follow, if the component set the attribute to hidden, browser does not display it anymore. How the developer can for instance have an error message which is displayed in grey and changes to red, etc? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh6se5vXNlLCPYC9Y8d:b-h4382b) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*
> I don't follow, if the component set the attribute to hidden, browser does not display it anymore. How the developer can for instance have an error message which is displayed in grey and changes to red, etc?

The `hidden` attribute is just a shortcut to `display: none;` (or so I believe), so you can override that with `display: block !important;` for example. Not the nicest, though, so I think it would be better to control the visiblity with CSS. And I prefer `display: none` over `visibility: hidden` so that it doesn’t reserve any space and need to be included in the browser’s layout phase.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

manolo:
Review status: 1 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh6vWS8D-C7qJUMcoPl:bkso3dr) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

> I don't follow, if the component set the attribute to hidden, browser does not display it anymore. How the developer can for instance have an error message which is displayed in grey and changes to red, etc?

The `hidden` attribute is just a shortcut to `display: none;` (or so I believe), so you can override that with `display: block !important;` for example. Not the nicest, though, so I think it would be better to control the visiblity with CSS. And I prefer `display: none` over `visibility: hidden` so that it doesn’t reserve any space and need to be included in the browser’s layout phase.
</blockquote></details>

Isn't `hidden` an HTML5 attribute which browsers natively handle? http://caniuse.com/#feat=hidden ?
IIUC, it behaves like `display:none` but you cannot do anything in CSS

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-Kh6x4fkkm9UIQK8WyxD:blcs44m) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*
> Also, the error message is currently announced even if it's hidden.

Should bind to `aria-hidden`. I don’t really like dom-if here either, as it does add more extra elements in the DOM vs just one single element.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
*[vaadin-text-field.html, line 98 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2BZhUnXt4l7tN4MPc:-Kh70EwOqlOCvdiJKQCh:b-87r54u) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b33ba1ede3d2484577767d5a6c8ed25c4e761b2b/vaadin-text-field.html#L98)):*

Still not convinced, why is better that is good that the apparition of the error message should affect the layout ? but the hidden does not reserve space. Both cases can be addressed by the developer since we provide the selector, so is a question of taste I guess.
</blockquote></details>

Like I said in the other thread, I prefer `display: none` instead of `visibility: hidden`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh72rJONEJJIQGs9I7t:b-f0uarx) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

Isn't `hidden` an HTML5 attribute which browsers natively handle? http://caniuse.com/#feat=hidden ?
IIUC, it behaves like `display:none` but you cannot do anything in CSS
</blockquote></details>

No. See http://jsbin.com/nuvuzidase/edit?html,css,output (tested in Chrome, Safari and Firefox).

You don’t even need `!important` to override it, apparently. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh73CwIirVJmgZXByw8:b-8bz3k8) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*
> You don’t even need !important to override it, apparently.

So I take my previous comment back, no need to do this with CSS, just bind to the `hidden` attribute IMO.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

manolo:
Review status: 1 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-Kh7328gX1TC1YuvLR6Z:bh6k99p) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*
> Also, the error message is currently announced even if it's hidden.

This is exactly the same than paper-input does which was the task premise for this

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

manolo:
Review status: 1 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh7bNG2kXf8jeRiEybR:b-idcp53) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*
> hidden=[[_hideError(errorMessage, invalid)]]

OK, let's go with the hidden approach, but I think we don't need any function this should be enough `hidden=[[!invalid]]`  @tomivirkki ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

tomivirkki:
*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh7gG5bP9hohCkymG0I:bjcw2as) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

> hidden=[[_hideError(errorMessage, invalid)]]

OK, let's go with the hidden approach, but I think we don't need any function this should be enough `hidden=[[!invalid]]`  @tomivirkki ?

</blockquote></details>

Yeh, that could be enough. Toggling the display will make the element jump if the message is not in an absolute container though. And if it's in an absolute container, it might overflow the component box, but I guess that's better than adding extra space for the error message below the component .

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

manolo:
Review status: 1 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-text-field.html, line 98 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2BZhUnXt4l7tN4MPc:-Kh7leo11Lj3zFENRahB:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b33ba1ede3d2484577767d5a6c8ed25c4e761b2b/vaadin-text-field.html#L98)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

Like I said in the other thread, I prefer `display: none` instead of `visibility: hidden`.
</blockquote></details>

Done.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh7lolvsoRNR9rJOeZy:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

Yeh, that could be enough. Toggling the display will make the element jump if the message is not in an absolute container though. And if it's in an absolute container, it might overflow the component box, but I guess that's better than adding extra space for the error message below the component .
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

tomivirkki:
*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh7tZt_HCY24oAEXPqu:b-tkxeaw) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*
> Toggling the display will make the element jump if the message is not in an absolute container though.

The error message should be in an absolute container. Currently it’s jumpy and the message still stretches the whole component: https://www.dropbox.com/s/r11ojllu3blk766/Kapture%202017-04-07%20at%2017.50.07.gif?dl=0

Try something like the following (these are not well tested, just to give an idea):
```
[part=error-message] {
  position: absolute;
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
  width: 100%;
  font-size: 0.7em; <- Maybe the raw version shouldn’t have this?
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

platosha:
:lgtm:

---

Reviewed 4 of 4 files at r6.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh7v1k2hKPXc3fWHx0h:b-626fq4) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

> Toggling the display will make the element jump if the message is not in an absolute container though.

The error message should be in an absolute container. Currently it’s jumpy and the message still stretches the whole component: https://www.dropbox.com/s/r11ojllu3blk766/Kapture%202017-04-07%20at%2017.50.07.gif?dl=0

Try something like the following (these are not well tested, just to give an idea):
```
[part=error-message] {
  position: absolute;
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
  width: 100%;
  font-size: 0.7em; <- Maybe the raw version shouldn’t have this?
}
```
</blockquote></details>

BTW Maybe the raw version shouldn’t have the absolute container either.

Limiting the error message to a single line is also controversial.

I would rather give an explicit size to the container instead, or even make it into a block element. Now that we have error messages of random length, it starts to make sense.

Note, that when text fields are inside the layout, the size is given by the layout, so that stretching would not happen.

cc @jouni 

---

*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-Kh7iObg_XZ9fh9_aHzN:b-bqsdfq) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*

> Also, the error message is currently announced even if it's hidden.

This is exactly the same than paper-input does which was the task premise for this
</blockquote></details>

BTW announcing the hidden message should not be a problem in dynamic use cases, when the error message is set together with the invalid state at the same time, as a result of validation.

Actually, announcing the hidden error message could be even a feature, giving hints on filling the field beforehand.

`<paper-input>`’s implementation looks like aiming at two use cases together, so that both dynamic error messages can give feedback after validation (they have `aria-live` on the error message for that), but also static error messages can give hints beforehand.

I think we don't need `aria-hidden` here, and don't need to remove the hidden error message announce. This should be thought out better, outside of this PR scope.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-:-Kh8-X8-l6N-P6J35MQy:bnfp4nl)*

manolo:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh802Dj1rqcMAxpt2no:bekhjh3) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

BTW Maybe the raw version shouldn’t have the absolute container either.

Limiting the error message to a single line is also controversial.

I would rather give an explicit size to the container instead, or even make it into a block element. Now that we have error messages of random length, it starts to make sense.

Note, that when text fields are inside the layout, the size is given by the layout, so that stretching would not happen.

cc @jouni 
</blockquote></details>

I also disagree of styling the error message.  Your suggestion implies that the position of the error message is out of the boundaries of the text-field which makes the component intrusive. It works in your case because the container of the div has enough space for the error, but try this simple demo, and you would get that the error overlaps other elements, or is not visible at all because it could be covered by other elements, (screenshot: http://g.recordit.co/iU0KANiqeD.gif )

```
  <div style="background: pink; padding: 2px">
    <vaadin-text-field required error-message="Error Message 1"></vaadin-text-field>
    <hr>
    <vaadin-text-field required error-message="Error Message 2"></vaadin-text-field>
    <br>
    <vaadin-text-field required error-message="Error Message 3"></vaadin-text-field>
  </div>
```

The `visibility: hidden` tried to avoid that problem because the layout always reserves space for the message.
But if we go with `display:none` we need either, give enough room in the element to place the error, or just make the error relayout the component when it appears.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-Kh8sYbNc9AFUZrOs3nr:b-yvaoza) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

I also disagree of styling the error message.  Your suggestion implies that the position of the error message is out of the boundaries of the text-field which makes the component intrusive. It works in your case because the container of the div has enough space for the error, but try this simple demo, and you would get that the error overlaps other elements, or is not visible at all because it could be covered by other elements, (screenshot: http://g.recordit.co/iU0KANiqeD.gif )

```
  <div style="background: pink; padding: 2px">
    <vaadin-text-field required error-message="Error Message 1"></vaadin-text-field>
    <hr>
    <vaadin-text-field required error-message="Error Message 2"></vaadin-text-field>
    <br>
    <vaadin-text-field required error-message="Error Message 3"></vaadin-text-field>
  </div>
```

The `visibility: hidden` tried to avoid that problem because the layout always reserves space for the message.
But if we go with `display:none` we need either, give enough room in the element to place the error, or just make the error relayout the component when it appears.



</blockquote></details>

No absolute positioning by default, no forcing to one line by default. Those are all up to the themes to decide. So by default, limit the width of the error message to the width of the field, and let the text wrap on as many lines as needed.

One more benefit of using `hidden` instead of `<dom-if>` is that we can then use CSS transitions to show/hide the message.

An example of both can be seen in this prototype of mine: http://codepen.io/jouni/full/dcrAu/ (type anything to the required fields, marked with a red dot, and blur the field to see the error message appear).

So it’s completely acceptable to let the error message cause a relayout, no need to reserve space for it in advance. A theme can decide to treat it differently if needed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-Kh8tMF0IaYptpolM91m:b-stmyqk) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*
> This is exactly the same than paper-input does which was the task premise for this

IMO, the error should only be announced when the field is invalid and the error message is visible. The could be a separate help text for the hints.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

tomivirkki:
*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhBKDVsGlg0S-XcSuO2:b7et0fv) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

No absolute positioning by default, no forcing to one line by default. Those are all up to the themes to decide. So by default, limit the width of the error message to the width of the field, and let the text wrap on as many lines as needed.

One more benefit of using `hidden` instead of `<dom-if>` is that we can then use CSS transitions to show/hide the message.

An example of both can be seen in this prototype of mine: http://codepen.io/jouni/full/dcrAu/ (type anything to the required fields, marked with a red dot, and blur the field to see the error message appear).

So it’s completely acceptable to let the error message cause a relayout, no need to reserve space for it in advance. A theme can decide to treat it differently if needed.
</blockquote></details>

If the error message shouldn't be absolute, how would you align the `<vaadin-text-field>`s in the following example: https://jsfiddle.net/w3omtang/

For reference I added `<paper-input>`s below. They use absolute with the error message and the component aligns well.

If the answer is "you're not supposed to use the raw version but always include a theme that takes care of the styling", then we should probably remove the whole style block from the raw version, right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

platosha:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhMGHwj5M8dy2jWkaJu:b9lm03d) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

If the error message shouldn't be absolute, how would you align the `<vaadin-text-field>`s in the following example: https://jsfiddle.net/w3omtang/

For reference I added `<paper-input>`s below. They use absolute with the error message and the component aligns well.

If the answer is "you're not supposed to use the raw version but always include a theme that takes care of the styling", then we should probably remove the whole style block from the raw version, right?
</blockquote></details>

BTW @tomi `align-items: baseline;` makes it look much better. https://jsfiddle.net/1kosqzor/

With `align-items: baseline;` and assuming that labels are either present or absent in all of the inputs, they are aligned well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

tomivirkki:
*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhMHPeJ_W7r9eOGFOj5:b6xobwp) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

BTW @tomi `align-items: baseline;` makes it look much better. https://jsfiddle.net/1kosqzor/

With `align-items: baseline;` and assuming that labels are either present or absent in all of the inputs, they are aligned well.
</blockquote></details>

That makes it broken by default. `<paper-input>` doesn’t have that kind of restriction and neither should `<vaadin-text-field>`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhMHzYO-gzE2Hq3EmB4:bgrtqrc) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

That makes it broken by default. `<paper-input>` doesn’t have that kind of restriction and neither should `<vaadin-text-field>`.
</blockquote></details>

I argue that Tomi’s use case is an edge case. In most cases, you have labels for all of your fields, or for none of them. And in case you have this situation, you can provide a label which is just one empty space character.
​

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

platosha:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhMHXQ7rF9nshKuG4xk:bb4rnnn) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

I argue that Tomi’s use case is an edge case. In most cases, you have labels for all of your fields, or for none of them. And in case you have this situation, you can provide a label which is just one empty space character.
​
</blockquote></details>

BTW @tomivirkki also we’re comparing raw `<vaadin-text-field>` with themed `<paper-input>` here. Material Design `<vaadin-text-field>` might (and probably will) have a label placeholder space regardless of the label presence, and therefore might be aligned.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

manolo:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhMH_mgAIiZfFcBkFYU:bjz4qck) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

BTW @tomivirkki also we’re comparing raw `<vaadin-text-field>` with themed `<paper-input>` here. Material Design `<vaadin-text-field>` might (and probably will) have a label placeholder space regardless of the label presence, and therefore might be aligned.
</blockquote></details>

I prefer not having any CSSing in the raw implementation. We already are trying not to style anything in the dom content, and that is the purpose of providing different themes for final users.

---

*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-KhMFvhA5-k0NTb_nT0V:b-pravic) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

> This is exactly the same than paper-input does which was the task premise for this

IMO, the error should only be announced when the field is invalid and the error message is visible. The could be a separate help text for the hints.
</blockquote></details>

There are different UX options for this. The one used here is the same paper-input uses and I think is safe in the sense that we don't change the UX users could be familiar with already. So as our component is accessible, and we don't need to demonstrate that our solution is better during the UX test phase.

To me probably the best solution would be an aria alert as is described here https://www.w3.org/WAI/GL/wiki/Using_ARIA_role_of_alert_for_Error_Feedback_in_Forms so as the user does not leave the field invalid without any feedback.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

tomivirkki:
*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhMJZpV1DeMk7U4xHdv:b-unyd7h) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

I prefer not having any CSSing in the raw implementation. We already are trying not to style anything in the dom content, and that is the purpose of providing different themes for final users.
</blockquote></details>

Why are we suddenly discussing the label here (I obviously shouldn’t have used them in the example), since it’s about the error message? Fixing broken layout shouldn’t be the theme’s responsibility.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

manolo:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhMJp9M8bqJod6OwYB1:b-2l57f6) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

Why are we suddenly discussing the label here (I obviously shouldn’t have used them in the example), since it’s about the error message? Fixing broken layout shouldn’t be the theme’s responsibility.
</blockquote></details>

What is broken layout for you, I dont see anything broken in this solution.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

tomivirkki:
*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhMMDwIG2a4uhlvbvgX:b-4tg0xb) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

What is broken layout for you, I dont see anything broken in this solution.
</blockquote></details>

Didn’t the jsfiddle example I sent you look broken? The problem is that the text field has different amount of spacing below depending on if it has an error message or not.

If we agree that the raw version can look broken by default and the themes should always contain “fixes” for the layout, then this is ok for me.

We could even consider removing other rules that aim to fix the appearance of the raw component (like label positioning) and leave more responsibility for the themes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

manolo:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhMQ5RHcVlHe0RwrFCU:b-29pwad) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*
> Didn’t the jsfiddle example I sent you look broken? 

I didn't see anything broken  If you remove all your styling and leave a raw demo layout behaves as expected. I don't see why removing the error message height should remain the UX is good (http://g.recordit.co/KqDRGIdNbc.gif) . If it's is needed to leave that space always I would go with the visibility solution, but that was already discused

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-KhMRsBFuEZkrejfu3Au:buakvgo) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*

There are different UX options for this. The one used here is the same paper-input uses and I think is safe in the sense that we don't change the UX users could be familiar with already. So as our component is accessible, and we don't need to demonstrate that our solution is better during the UX test phase.

To me probably the best solution would be an aria alert as is described here https://www.w3.org/WAI/GL/wiki/Using_ARIA_role_of_alert_for_Error_Feedback_in_Forms so as the user does not leave the field invalid without any feedback.
</blockquote></details>

There’s probably some difference in how we see the error-message attribute/property being used. I see it as a way to specify the field validation error message in advance declaratively so that you don’t need to update that programmatically when the validation is actually done. Consider the following example:

```html
<vaadin-text-field label="Label" required error-message="You must provide a value for this field"></vaadin-text-field>
```

When the user focuses the field the first time, what should be announced? I can see two options:
- “field, required, Label"
- "field, required, Label, You must provide a value for this field"

The former seems like it’s enough. The latter doesn’t seem wrong either, but somehow redundant. Also, it might be wrong depending on the error message phrasing, for example, if it announced: “You did not provide a valid email address”.

So, I was thinking, the error message should only be announced when the field is actually invalid, as it aligns with the visual representation as well. The paper-input implementation seems inconsistent in that regard, if the error message is indeed announced even before the field is marked as invalid, and I would not hold that up as a perfect example.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

tomivirkki:
:lgtm:

---

Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhMUNe3mROEHPY6HoLy:bqd8iqu) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

> Didn’t the jsfiddle example I sent you look broken? 

I didn't see anything broken  If you remove all your styling and leave a raw demo layout behaves as expected. I don't see why removing the error message height should remain the UX is good (http://g.recordit.co/KqDRGIdNbc.gif) . If it's is needed to leave that space always I would go with the visibility solution, but that was already discused
</blockquote></details>

If you don’t use flex (like in you gif) or always provide the label (in flex) then it aligns ok. Discussed with Jouni and he’s ok with leaving the responsibility to individual themes/developer

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-:-KhMUugTcILguQ3j9oMG:bnfp4nl)*

tomivirkki:
Reviewed 1 of 4 files at r1, 4 of 4 files at r6.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

demo/validators.html
```diff
@@ -113,7 +113,9 @@
 
                 this.ajax = document.getElementById('ajax1');
                 this.ajax.addEventListener('response', function() {
-                  this.invalid = this.ajax.lastResponse.elements.indexOf(this.value) < 0;
+                  var elementName = this.value || '';
+                  this.invalid = this.ajax.lastResponse.elements.indexOf(elementName.toLowerCase()) < 0;
+                  this.errorMessage = this.invalid ? '"' + elementName + '" is not a Chemical Element' : '';
```
jouni: This is exactly not how I see this attribute/property being used. The message should be set in advance declaratively, not programmatically after validation.

vaadin-text-field.html
```diff
@@ -69,6 +69,10 @@
       .label-placeholder[no-label] {
         display: none;
       }
+
+      [part=error-message]:empty {
```
jouni: Considering the use case that the error message is provided declaratively in advance, this CSS rule does not work (as the message is not empty).

manolo:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-KhMXYED2i8AAk32Ejrz:b-psdqnw) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

There’s probably some difference in how we see the error-message attribute/property being used. I see it as a way to specify the field validation error message in advance declaratively so that you don’t need to update that programmatically when the validation is actually done. Consider the following example:

```html
<vaadin-text-field label="Label" required error-message="You must provide a value for this field"></vaadin-text-field>
```

When the user focuses the field the first time, what should be announced? I can see two options:
- “field, required, Label"
- "field, required, Label, You must provide a value for this field"

The former seems like it’s enough. The latter doesn’t seem wrong either, but somehow redundant. Also, it might be wrong depending on the error message phrasing, for example, if it announced: “You did not provide a valid email address”.

So, I was thinking, the error message should only be announced when the field is actually invalid, as it aligns with the visual representation as well. The paper-input implementation seems inconsistent in that regard, if the error message is indeed announced even before the field is marked as invalid, and I would not hold that up as a perfect example.
</blockquote></details>

The premise of this PR is to take paper-input as inspiration and do not complicate the solution, this is also the reason of not making higher estimation than 2 points. 

I absolutely agree that this is not the perfect UX, but anyways there is a story (#48) to add accessibility where probably we can face all options. Or maybe a separated issue and PR.

  

---

*[vaadin-text-field.html, line 73 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh1zmWetvjQntcuUMny-r6-73:-KhMYrDeneRDW_7KTK5w:b-ym08n5) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/8c989789b144517e660ac26a68f3684ee4aac491/vaadin-text-field.html#L73)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

Considering the use case that the error message is provided declaratively in advance, this CSS rule does not work (as the message is not empty).
</blockquote></details>

I don't follow, we need this when there is no error message provided or we remove it imperatively 

---

*[demo/validators.html, line 118 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh6Vo2QD_coLcxwpb4O-r6-118:-KhMX9TllNexxWejmIkf:b-rqq62a) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/8c989789b144517e660ac26a68f3684ee4aac491/demo/validators.html#L118)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

This is exactly not how I see this attribute/property being used. The message should be set in advance declaratively, not programmatically after validation.
</blockquote></details>

This is a suggestion from @platosha to demonstrate that the attribute could be used imperatively. I don't have a strong preference of having it here or not, but was added during the review process.

BTW. This case also demonstrate that the user can easily offer a solution of not announcing the error message based on the invalidity of the control.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-KhMbJ9kW8me4JpGy8Xj:bimzxv6) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*

The premise of this PR is to take paper-input as inspiration and do not complicate the solution, this is also the reason of not making higher estimation than 2 points. 

I absolutely agree that this is not the perfect UX, but anyways there is a story (#48) to add accessibility where probably we can face all options. Or maybe a separated issue and PR.

  
</blockquote></details>

Yeah, sorry about making this (again) more complex than perhaps necessary, and not having done these discussions before and having a clearer spec. Largely my fault. I also tend to fall into perfectionism really easily.

So, as agreed in Hangout, just change it so that the error message is only announced for invalid fields and we are good to go!

---

*[vaadin-text-field.html, line 73 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh1zmWetvjQntcuUMny-r6-73:-KhMcGPhFVpZTCUsEHcJ:b-e36f6l) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/8c989789b144517e660ac26a68f3684ee4aac491/vaadin-text-field.html#L73)):*

I don't follow, we need this when there is no error message provided or we remove it imperatively 
</blockquote></details>

Ah, sorry, I forgot there’s the `hidden="[[!invalid]]"`binding making sure the error message is not shown for valid fields. In that case this rule is redundant and can be removed completely.

---

*[demo/validators.html, line 118 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh6Vo2QD_coLcxwpb4O-r6-118:-KhMbWgzxsVX-ZteoGjd:b76a1h4) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/8c989789b144517e660ac26a68f3684ee4aac491/demo/validators.html#L118)):*

This is a suggestion from @platosha to demonstrate that the attribute could be used imperatively. I don't have a strong preference of having it here or not, but was added during the review process.

BTW. This case also demonstrate that the user can easily offer a solution of not announcing the error message based on the invalidity of the control.
</blockquote></details>

Yeah, this is a good example to have I agree. Apologies for the wording. But could we add another example where the error message is added declaratively, as it is something we are now going to support, by only announcing the error message when the field is actually invalid. As discussed with Anton, we plan to add a “hint/help text” property in the future, which would serve the purpose of guiding the user before any validation is done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

manolo:
Review status: 1 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[vaadin-text-field.html, line 73 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh1zmWetvjQntcuUMny-r6-73:-KhNBp3nDy6heLftvu2F:bepdd1u) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/8c989789b144517e660ac26a68f3684ee4aac491/vaadin-text-field.html#L73)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

Ah, sorry, I forgot there’s the `hidden="[[!invalid]]"`binding making sure the error message is not shown for valid fields. In that case this rule is redundant and can be removed completely.
</blockquote></details>

Well, it was no redundant, you needed it in order to cover the case when invalid is true and there is no message.

---

*[demo/validators.html, line 118 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh6Vo2QD_coLcxwpb4O-r6-118:-KhO1vF2im2OvI7bc8vy:b670bca) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/8c989789b144517e660ac26a68f3684ee4aac491/demo/validators.html#L118)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

Yeah, this is a good example to have I agree. Apologies for the wording. But could we add another example where the error message is added declaratively, as it is something we are now going to support, by only announcing the error message when the field is actually invalid. As discussed with Anton, we plan to add a “hint/help text” property in the future, which would serve the purpose of guiding the user before any validation is done.
</blockquote></details>

With latest code error is only announced in case invalid is true, so no need to add another error since in the index.html we have one covering this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

jouni:
:lgtm:

---

Review status: 1 of 5 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-:-KhO4CVW9Anx2D8HwHiX:bnfp4nl)*

manolo:
Review status: 1 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-text-field.html, line 77 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh5gxC5DFHL-79o-2yD:-KhO5bGQPu0Xax3qN5jM:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L77)):*

If you don’t use flex (like in you gif) or always provide the label (in flex) then it aligns ok. Discussed with Jouni and he’s ok with leaving the responsibility to individual themes/developer
</blockquote></details>

Done.

---

*[vaadin-text-field.html, line 115 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56#-Kh2mfZCczDp17YhAhli:-KhO5acdCHckoNiyD98U:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/9ec4effa947d633ce63d2d7c76febf9722d3ea36/vaadin-text-field.html#L115)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

Yeah, sorry about making this (again) more complex than perhaps necessary, and not having done these discussions before and having a clearer spec. Largely my fault. I also tend to fall into perfectionism really easily.

So, as agreed in Hangout, just change it so that the error message is only announced for invalid fields and we are good to go!
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/56)*

PR #69, opened 2017-04-19
Improve accessibility

### LINKED ISSUE #48: Accessibility support

tomivirkki:
Label announced. Aria attributes for readonly, invalid and required added.

Closes #48

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/69)


manolo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-text-field.html, line 316 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/69#-Ki4jAZOldFq2Ys0GOKh:-Ki4iYflVa4vE0vQTf84:b-ga3815) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/d3a09cfaf7c2114479c718170560b32967b92756/vaadin-text-field.html#L316)):*
> ```HTML
>           }
> 
>           var uniqueId = TextFieldElement._uniqueId = 1 + TextFieldElement._uniqueId || 0;
> ```

I would maintain the comment, so as the reader knows why, and also in order to remove this hack when there shadydom support disappears. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/69)*

tomivirkki:
*[vaadin-text-field.html, line 316 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/69#-Ki4jAZOldFq2Ys0GOKh:-Ki4kG7XO_OLevFqzhlb:br8t8yo) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/d3a09cfaf7c2114479c718170560b32967b92756/vaadin-text-field.html#L316)):*

I would maintain the comment, so as the reader knows why, and also in order to remove this hack when there shadydom support disappears. 
</blockquote></details>

I didn't just remove the comment but also removed the whole check for ShadyDOM. There's no reason to make this work differently in Shadow/Shady. Don't expect us to be able to drop shadydom support since we support IE11.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/69)*

manolo:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-text-field.html, line 316 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/69#-Ki4jAZOldFq2Ys0GOKh:-Ki8qD-sKQbTquY4jh4a:b-nkh2z) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/d3a09cfaf7c2114479c718170560b32967b92756/vaadin-text-field.html#L316)):*

I didn't just remove the comment but also removed the whole check for ShadyDOM. There's no reason to make this work differently in Shadow/Shady. Don't expect us to be able to drop shadydom support since we support IE11.
</blockquote></details>

Yes the removal of the check is ok.
What I meant is that the uniqueId calculation is not needed when everything is shadow, and it's here exclusively because of shady.
Polymer announced that all polyfill will be removed when all browsers support specs, I don't expect to remove support for IE11 in brief, but I believe we can do in a long term.
I don't think leaving the comment is bad anyways. It will help (to me) when reading back this code in some months. Not leaving this comment as a blocker though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/69)*

manolo:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/69#-:-Ki9BBfGEGgpDIWPwUe1:bnfp4nl)*

platosha:
:lgtm:

---

Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/69#-:-KiEL8V4NgK6RLqHBYeA:bnfp4nl)*

manolo:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/69)*

PR #98, opened 2017-06-08
Implement the vaadin-password-field

limonte:
Implements the first item of https://github.com/vaadin/bakery-app-starter-polymer/issues/22

> New element inside the vaadin-text-field repository

This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/98)


manolo:
Need some tests

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-text-field.html, line 315 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Km6UOuBQFdqk9O5bQ2E:-Km6UOuBQFdqk9O5bQ2F:b-xp520v) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/97e64d122f7306e4f39a68209feb3812c08305a4/vaadin-text-field.html#L315)):*
> ```HTML
>               type: Boolean
>             },
> ```

There is no need to modify this component. We are kind of modifying the API so as user can to modify the behavior by extending. If we want to encourage the user to extend this, we should do via a public method for setting the type and checking what we support.

---

*[demo/index.html, line 20 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Km6V4QQodPXXZlvMXvF:-Km6V4QQodPXXZlvMXvG:bzgutln) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/97e64d122f7306e4f39a68209feb3812c08305a4/demo/index.html#L20)):*
> ```HTML
>     <demo-navigation></demo-navigation>
> 
>     <h3>Input types</h3>
> ```

Not clear to me. I would have `Password Element` and remove the normal text field. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Km6VElDu09OUr3WiEIj:b-3zjy12)*

limonte:
Review status: 0 of 13 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-text-field.html, line 315 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Km6UOuBQFdqk9O5bQ2E:-Km_cvtwShwPDZxI61ro:bcewtb4) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/97e64d122f7306e4f39a68209feb3812c08305a4/vaadin-text-field.html#L315)):*

There is no need to modify this component. We are kind of modifying the API so as user can to modify the behavior by extending. If we want to encourage the user to extend this, we should do via a public method for setting the type and checking what we support.
</blockquote></details>

Did this via Sauli's suggestion from [this thread](https://www.flowdock.com/app/vaadin/elements-team/threads/1PcMgpl-uaqP-dSHZHcS97W0_g8?filter=search&query=as%20simple) by simply setting input type 

---

*[demo/index.html, line 20 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Km6V4QQodPXXZlvMXvF:-Km_dKY3fSdd6DA9W0WA:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/97e64d122f7306e4f39a68209feb3812c08305a4/demo/index.html#L20)):*

Not clear to me. I would have `Password Element` and remove the normal text field. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

tomivirkki:
*[vaadin-text-field.html, line 315 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Km6UOuBQFdqk9O5bQ2E:-Km_hdHu_lK2Syfb2W9A:bd0iifg) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/97e64d122f7306e4f39a68209feb3812c08305a4/vaadin-text-field.html#L315)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

Did this via Sauli's suggestion from [this thread](https://www.flowdock.com/app/vaadin/elements-team/threads/1PcMgpl-uaqP-dSHZHcS97W0_g8?filter=search&query=as%20simple) by simply setting input type 
</blockquote></details>

One option to consider is to modify the superclass template instead. That way the modifications aren’t done for individual component instances but only once for the static template during registration:

```
static get template() {
 // Change the input type
 const tpl = super.template.cloneNode(true);
 tpl.content.getElementById('input').setAttribute('type', 'password');

 // Append the eye-con
 const icon = document.createElement('div');
 icon.innerText = 'EYE';
 tpl.content.querySelector('[part="input-field"]').appendChild(icon);

 return tpl;
}
```

Documentation for this: https://www.polymer-project.org/2.0/docs/devguide/dom-template#inherited-templates

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-text-field.html, line 315 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Km6UOuBQFdqk9O5bQ2E:-KmaGH3r8nzR2ufzfsFD:bf6nwzz) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/97e64d122f7306e4f39a68209feb3812c08305a4/vaadin-text-field.html#L315)):*

One option to consider is to modify the superclass template instead. That way the modifications aren’t done for individual component instances but only once for the static template during registration:

```
static get template() {
 // Change the input type
 const tpl = super.template.cloneNode(true);
 tpl.content.getElementById('input').setAttribute('type', 'password');

 // Append the eye-con
 const icon = document.createElement('div');
 icon.innerText = 'EYE';
 tpl.content.querySelector('[part="input-field"]').appendChild(icon);

 return tpl;
}
```

Documentation for this: https://www.polymer-project.org/2.0/docs/devguide/dom-template#inherited-templates
</blockquote></details>

Thanks a lot for this link! Will go with the template inheritance.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 0 of 15 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-text-field.html, line 315 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Km6UOuBQFdqk9O5bQ2E:-KmexzWZRDpnft1GTDnp:b-2n4ytq) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/97e64d122f7306e4f39a68209feb3812c08305a4/vaadin-text-field.html#L315)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

Thanks a lot for this link! Will go with the template inheritance.
</blockquote></details>

Done, take a look.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

tomivirkki:
*[vaadin-password-field.html, line 37 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-KmeyorOsu1p1IuZjinc:-KmeyorOsu1p1IuZjind:b4rxs7m) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/46ca76d167d8197d53574a890539ddf69e512e9f/vaadin-password-field.html#L37)):*
> ```HTML
>        * See vaadin-text-field.html for the styling documentation
>        *
>        * In addition to vaadin-text-field parts, here's the list of vaadin-text-field specific parts
> ```

Typo: "vaadin-text-field" -> "vaadin-password-field"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 0 of 15 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-password-field.html, line 37 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-KmeyorOsu1p1IuZjinc:-KmfG0Asa_K2b86sInT0:b-3eywoa) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/46ca76d167d8197d53574a890539ddf69e512e9f/vaadin-password-field.html#L37)):*

Typo: "vaadin-text-field" -> "vaadin-password-field"
</blockquote></details>

oops, fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

manolo:
Reviewed 15 of 17 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-password-field.html, line 15 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kmv2fnCYjbVYzb7eSlA:-Kmv2fnCYjbVYzb7eSlB:bxk2yy4) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/6ebf5142509f463aec334fe3016370345d564d91/vaadin-password-field.html#L15)):*
> ```HTML
>     <style>
>       [part="eye-icon"]::before {
>         content: '👁';
> ```

I think the Icon is not noticeable, It might be design of the icon, or its size, or the transparence that it does not look a proper eye.

We could provide our own eye icons based on vaadin-icons



Also it should be a 2 states when showing password and when its not showing it like in this example



---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 9 of 17 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[vaadin-password-field.html, line 15 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kmv2fnCYjbVYzb7eSlA:-KmzgWqB4x1E3v8g5reE:bfl5lsu) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/6ebf5142509f463aec334fe3016370345d564d91/vaadin-password-field.html#L15)):*

I think the Icon is not noticeable, It might be design of the icon, or its size, or the transparence that it does not look a proper eye.

We could provide our own eye icons based on vaadin-icons



Also it should be a 2 states when showing password and when its not showing it like in this example



</blockquote></details>

Good point! Done, take a look.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

manolo:
Review status: 9 of 18 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[vaadin-password-field.html, line 27 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kmzg8T4kBgiFGjZ4FBA:-Kn-7Z0xhjrKwrMeFdVo:b-n3pzu3) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/3bf121bb460c7ce49a5cc20fe7ad3391a27a71bb/vaadin-password-field.html#L27)):*

We should probably align how our elements define clickable icons (iron-icon vs font icons). Combo-box and date-picker use font icons and I'd personally prefer that approach since it allows you to modify the icons in the theme.
</blockquote></details>

+1 for font icons

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 9 of 16 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-password-field.html, line 27 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kmzg8T4kBgiFGjZ4FBA:-Kn-bcvwhamLmhfOPEa4:b4497ey) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/3bf121bb460c7ce49a5cc20fe7ad3391a27a71bb/vaadin-password-field.html#L27)):*

+1 for font icons
</blockquote></details>

Done, thanks for the tip. Font icons are much better with the LightHouse performance test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

tomivirkki:
Toggling the visibility feels flickery since the outline disappears while mouse button is pressed. Can something be done to fix this? I noticed it's a common problem with all prefixes/suffixes however.

The icon is quite hard to hit with mobile device (combo-box and date-picker icons assumably have the same issue).

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Kn4eWpkkbkv27pILbwB:b-2ejztt)*

limonte:
> Toggling the visibility feels flickery since the outline disappears while mouse button is pressed. Can something be done to fix this? I noticed it's a common problem with all prefixes/suffixes however.

Probably this is the control-state-mixin issue, will take a look.

> The icon is quite hard to hit with mobile device (combo-box and date-picker icons assumably have the same issue).

Good catch, will debug it.

---

Review status: 9 of 16 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Kn4r8U4JLCk7hRneoMJ:b-2j73l4)*

Saulis:
Question: is there a way to toggle the visibility with keyboard? Should there be? Anybody seen any examples of this on the internets?

---

Review status: 9 of 16 files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-password-field.html, line 23 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn4zO5PBlbO-vZcANu6:-Kn4zO5PBlbO-vZcANu7:b-70mwnn) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L23)):*
> ```HTML
>   <template>
>     <style>
>       [part="eye-icon-show"],
> ```

Sorry if this has been discussed already before – could we have a more functional name for the part/icon element, since it might not be an "eye" icon always. Maybe `show-password-icon` and `hide-password-icon` ?

---

*[vaadin-password-field.html, line 121 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn5-HHxmVP_yu8kn8z2:-Kn5-HHxmVP_yu8kn8z3:b73u5yc) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L121)):*
> ```HTML
>         ready() {
>           super.ready();
>           this.passwordVisible = false;
> ```

This shouldn't be needed since `passwordVisible´ already has a default value set?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Kn5-s2-26fcPFn0tiIa:bll4rfm)*

limonte:
> Question: is there a way to toggle the visibility with keyboard? 

No.

> Should there be? 

I don't think so.

> Anybody seen any examples of this on the internets?

https://mail.protonmail.com/login 

---

Review status: 9 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-password-field.html, line 23 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn4zO5PBlbO-vZcANu6:-Kn5Be64nxTB7OzcVDkV:b-ey9kbu) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L23)):*

Sorry if this has been discussed already before – could we have a more functional name for the part/icon element, since it might not be an "eye" icon always. Maybe `show-password-icon` and `hide-password-icon` ?
</blockquote></details>

Nice suggestion, done.

---

*[vaadin-password-field.html, line 121 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn5-HHxmVP_yu8kn8z2:-Kn5BhrAt7BhOdCRYFhm:b-9yy7en) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L121)):*

This shouldn't be needed since `passwordVisible´ already has a default value set?
</blockquote></details>

Right, done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Kn5BlUalzTJsf9Qw9-d:b-tbaoli)*

manolo:
Review status: 9 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-password-field.html, line 27 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn5AAInP9WsT0MzdGqo:-Kn5AAInP9WsT0MzdGqp:btza2gc) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L27)):*
> ```HTML
>         font-family: 'vaadin-password-field-icons';
>         cursor: pointer;
>       }
> ```

Why do we have two buttons instead of one? I see simpler to have a css rule based on the `passwordVisible` attribute

```
[part="toggle-button"]:before {
   content: "\e900";
}

:host([password-visible]) [part="toggle-button"]:before {
   content: "\e901";
}

```

---

*[vaadin-password-field.html, line 129 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn5CLmq8O-vULS2el1J:-Kn5CLmq8O-vULS2el1K:b-vo79py) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L129)):*
> ```HTML
>           } else {
>             this.$.input.type = 'password';
>           }
> ```

Use the ternary operator here, that makes code simpler in this case.

---

*[vaadin-password-field.html, line 132 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn5BhNk--hYs_sckggC:-Kn5BhNk--hYs_sckggD:bwyb8hw) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L132)):*
> ```HTML
>         }
> 
>         _showPassword() {
> ```

having just one button as mentioned you only have to maintain one method and less code

```
_togglePassword() {
  this.passwordVisible = !this.passwordVisible;
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

manolo:
Since eye icon is a native feature in IE you should either disable this implementation for these browsers, otherwise you see both icons at the same time with weird behaviour.  Otherwise you must hide the icons in IE using CSS

---

Review status: 9 of 16 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Kn5JmHRNPLyt1ElEj8a:b-tus9wx)*

limonte:
Review status: 9 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-password-field.html, line 27 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn5AAInP9WsT0MzdGqo:-Kn8bq3WAvZo3DvrmdsJ:b-uckx6c) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L27)):*

Why do we have two buttons instead of one? I see simpler to have a css rule based on the `passwordVisible` attribute

```
[part="toggle-button"]:before {
   content: "\e900";
}

:host([password-visible]) [part="toggle-button"]:before {
   content: "\e901";
}

```
</blockquote></details>

I was thinking that two icons will be better for theming and customization, but you are right. Done.

---

*[vaadin-password-field.html, line 129 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn5CLmq8O-vULS2el1J:-Kn8c47QI8jq-q7yTqc2:b-xmdfr) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L129)):*

Use the ternary operator here, that makes code simpler in this case.
</blockquote></details>

I will stick to old good `if` for since I have added `togglePasswordTitle` here. 

---

*[vaadin-password-field.html, line 132 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn5BhNk--hYs_sckggC:-Kn8c2XCUfbWbMX1zEFR:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L132)):*

having just one button as mentioned you only have to maintain one method and less code

```
_togglePassword() {
  this.passwordVisible = !this.passwordVisible;
}
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Neither I nor gemini can't see any native eye icon in IE.

---

Review status: 9 of 16 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Kn8fekdlQiNgPqU5Xfr:bm27ta5)*

limonte:
> Anybody seen any examples of this on the internets?

Another example is new Polymer-based UI of chrome://settings/passwords User is able to toggle the password visibility form the keyboard there. But their eye looks like a standalone component out of scope of the password field.

---

Review status: 9 of 16 files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Kn8gNCbGeQYUiYVWbki:bn7qkwn)*

manolo:
> Neither I nor gemini can't see any native eye icon in IE.

I don't know about gemini, but just manually if you start typing the issue is in both Edge and IE11

http://g.recordit.co/Jy3X2JS7b0.gif

---

Review status: 9 of 16 files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Kn8h57gVh9PUilrE8rO:b-knlfis)*

manolo:
Review status: 9 of 16 files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-password-field.html, line 129 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn5CLmq8O-vULS2el1J:-Kn8hup7cZDsTp8UGRCJ:b-ijtqev) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L129)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

I will stick to old good `if` for since I have added `togglePasswordTitle` here. 
</blockquote></details>

I'm against adding the tittle here unless it's i18n-able 

---

*[vaadin-password-field.html, line 81 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8i7hnH8NblKKYe6iE:-Kn8i7hoRl9GgkpIlQA2:bxmpqv3) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b5647a2a0e60f27987bd317ec1e5ca73efb8b481/vaadin-password-field.html#L81)):*
> ```HTML
>              * True if the password is visible ([type=text]).
>              */
>             passwordVisible: {
> ```

Thinking more about the API we should offer, I think we should behave like native implementation does in IE, so IMO

1. passwordVisible should be read-only, so as the developer does not show the password visible by default. It must be the user who push the button to see the password when he knows anyone around could be looking at the screen.
2. Developer should have a way to disable this feature at all. In IE you have a CSS property for that, here we can use the theme, but I think it should be better to have a flag attribute for this. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

tomivirkki:
Review status: 9 of 16 files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-password-field.html, line 23 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn4zO5PBlbO-vZcANu6:-Kn8j48LwSjjYIRXqi4u:bm8ii7) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L23)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

Nice suggestion, done.
</blockquote></details>

The naming of these clickable icons should be aligned between other Vaadin Elements. For example date-picker has "toggle-button" (not "-icon" or "toggle-dropdown-button")

---

*[vaadin-password-field.html, line 25 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8hmev6O-0egtC_JeQ:-Kn8hmev6O-0egtC_JeR:bb5uuzj) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L25)):*
> ```HTML
>       [part="toggle-password-icon"] {
>         font-family: 'vaadin-password-field-icons';
>         cursor: pointer;
> ```

This is a non-functional style so it should be placed in a "vaadin-password-field-default-theme" -module

---

*[vaadin-password-field.html, line 38 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8ffaXHlt9rpzRza5h:-Kn8ffaXHlt9rpzRza5i:bu6spaf) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L38)):*
> ```HTML
> 
>     <div
>       id="togglePasswordIcon"
> ```

As this id is not used by the component in any way, it's practically a test helper. I'd prefer having only the necessary stuff in our components. Not a blocker.

---

*[vaadin-password-field.html, line 71 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8i5I0gwUnZKuN4-An:-Kn8i5I0gwUnZKuN4-Ao:bjtid49) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L71)):*
> ```HTML
>       let memoizedTemplate;
> 
>       class PasswordFieldElement extends Polymer.GestureEventListeners(Vaadin.TextFieldElement) {
> ```

This component extension expects that the super component extends VaadinThemableMixin instead of extending it by itself. To be 100% clean, password field should also import and use the mixin because it declares themable parts. 

---

*[vaadin-password-field.html, line 111 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8fp9RTXBFyxb_gEP5:-Kn8fp9RTXBFyxb_gEP6:brdmrj3) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L111)):*
> ```HTML
>           if (passwordVisible) {
>             this.$.input.type = 'text';
>             this.togglePasswordTitle = 'Hide password';
> ```

We can't make an assumption that the application UI is in english. I'd just leave the tooltip feature out for now and revisit it later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

manolo:
Review status: 9 of 16 files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-password-field.html, line 81 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8i7hnH8NblKKYe6iE:-Kn8jCJfeShekJYLKF-0:b-snwaiq) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b5647a2a0e60f27987bd317ec1e5ca73efb8b481/vaadin-password-field.html#L81)):*

Thinking more about the API we should offer, I think we should behave like native implementation does in IE, so IMO

1. passwordVisible should be read-only, so as the developer does not show the password visible by default. It must be the user who push the button to see the password when he knows anyone around could be looking at the screen.
2. Developer should have a way to disable this feature at all. In IE you have a CSS property for that, here we can use the theme, but I think it should be better to have a flag attribute for this. 
</blockquote></details>

Additionally, if we set passwordVisible as readOnly and default valued to false, we don't need observers at all because the toggle method could do all the work instead

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
I can't reproduce this, different version maybe? But I added `::-ms-reveal`, could you please test it out?

---

Review status: 9 of 16 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Kn8pJm0fRgnt84hQf6F:bt4c14l)*

limonte:
Review status: 8 of 16 files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-password-field.html, line 23 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn4zO5PBlbO-vZcANu6:-Kn92X-OzpiP6K9Z6DBb:be1fd64) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L23)):*

The naming of these clickable icons should be aligned between other Vaadin Elements. For example date-picker has "toggle-button" (not "-icon" or "toggle-dropdown-button")
</blockquote></details>

Makes sense, renamed to `toggle-button`

---

*[vaadin-password-field.html, line 129 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn5CLmq8O-vULS2el1J:-Kn92Kn7k-m1lfpmHciv:b-i5g3uh) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b2dab41ece177c5b500eafbf1f152e22683ec8a2/vaadin-password-field.html#L129)):*

I'm against adding the tittle here unless it's i18n-able 
</blockquote></details>

Removed tooltip for now. Ternary operator is now used. 

---

*[vaadin-password-field.html, line 38 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8ffaXHlt9rpzRza5h:-Kn92UoDkRPZ5i6_SawP:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L38)):*

As this id is not used by the component in any way, it's practically a test helper. I'd prefer having only the necessary stuff in our components. Not a blocker.
</blockquote></details>

Done.

---

*[vaadin-password-field.html, line 111 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8fp9RTXBFyxb_gEP5:-Kn92HkA2Dj9fk-DiEoY:b-baag0f) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L111)):*

We can't make an assumption that the application UI is in english. I'd just leave the tooltip feature out for now and revisit it later.
</blockquote></details>

Removed tooltip for now.

---

*[vaadin-password-field.html, line 81 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8i7hnH8NblKKYe6iE:-Kn923-_pRkhmg9Y0veb:bn3uvb5) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/b5647a2a0e60f27987bd317ec1e5ca73efb8b481/vaadin-password-field.html#L81)):*
> passwordVisible should be read-only

Done.

> Developer should have a way to disable this feature at all.

Done with the `hide-toggle-button` attribute

> Additionally, if we set passwordVisible as readOnly and default valued to false, we don't need observers at all because the toggle method could do all the work instead

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

manolo:
I don't think replacing the native implementation with synthetic one is the best option here, remember #usetheplatform when available. Just figure out when the native is implemented and set the hideToggleButton to true.

---

Review status: 8 of 16 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-password-field.html, line 38 at r8](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn9Alc5ujiWFt8wKn_u:-Kn9Alc5ujiWFt8wKn_v:bm2lv8e) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/e3eee66d20895b49e448384e40e589c083fa098f/vaadin-password-field.html#L38)):*
> ```HTML
>       /* Hide the native eye icon for IE/Edge */
>       ::-ms-reveal {
>         display: none;
> ```

remove this in favour of hideToggleButton = false when Edge and Trident

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Kn99kCfesMmEBTjHHBT:bmv6xix)*

limonte:
Review status: 8 of 16 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-password-field.html, line 25 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8hmev6O-0egtC_JeQ:-Kn93AfaBdzO3B6B2-aM:bxtrnj9) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L25)):*

This is a non-functional style so it should be placed in a "vaadin-password-field-default-theme" -module
</blockquote></details>

I can't get `vaadin-password-field-default-theme` work. Probably that's because of inheritance :(

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 8 of 16 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-password-field.html, line 38 at r8](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn9Alc5ujiWFt8wKn_u:-Kn9K7_feU070ItF7h0O:btcqtg7) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/e3eee66d20895b49e448384e40e589c083fa098f/vaadin-password-field.html#L38)):*

remove this in favour of hideToggleButton = false when Edge and Trident
</blockquote></details>

Detecting browsers (not feature) is bad. I try to hide the native IE/Edge eye icon instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 8 of 16 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-password-field.html, line 71 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8i5I0gwUnZKuN4-An:-Kn9KT-c6B5a4wR0S3lF:b-ne008i) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L71)):*

This component extension expects that the super component extends VaadinThemableMixin instead of extending it by itself. To be 100% clean, password field should also import and use the mixin because it declares themable parts. 
</blockquote></details>

I'm not sure I understand this comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

manolo:
Review status: 8 of 16 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-password-field.html, line 38 at r8](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn9Alc5ujiWFt8wKn_u:-Kn9MMbb2MUe7z6kPchg:b-s16lma) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/e3eee66d20895b49e448384e40e589c083fa098f/vaadin-password-field.html#L38)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

Detecting browsers (not feature) is bad. I try to hide the native IE/Edge eye icon instead.
</blockquote></details>

In this case this assumption does not work. Eye password feature is implemented in all IE11 and Edge versions, so detecting that the browser it that is more than enough. There is no way to detect the feature due to the lack of an API, but trying to hack whether the `::-ms-reveal` is effective by adding it to the stylesheet and testing whether it's accepted in the browser, I don't know if that is going to work. So instead of hacks for the feature, just a simple  UA regexp is enough (`/Trident|Edge/.test(ua)`) . We already use this technique for disabling/enabling features in other cases. I seriously doubt that we would penalise any other browser by using this check. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

tomivirkki:
Review status: 8 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-password-field.html, line 25 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8hmev6O-0egtC_JeQ:-Kn9S9C6uSKbNFOfkRS1:bthwmpj) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L25)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

I can't get `vaadin-password-field-default-theme` work. Probably that's because of inheritance :(
</blockquote></details>

Ok, debugged this a little. The error in the current implementation is that VaadinTextField's template() function is never called when registering the pwf (the inheritance doesn't work as expected). Meaning that you should instead:

1. get text-field's template with `super.template.cloneNode(true)` and store it as the `memoizedTemplate` (you must name the pwf dom-module as something else that "vaadin-password-field" or else `super.template` will fetch that instead)
1. Get pwf template with Polymer.DomModule.import
1. Append the the "toggle-button" and `<style>`from pwf template to the `memoizedTemplate`

---

*[vaadin-password-field.html, line 71 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8i5I0gwUnZKuN4-An:-Kn9P3qdeeYwXAwi6Fej:bckdato) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L71)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

I'm not sure I understand this comment.
</blockquote></details>

Meaning that instead of relying that `<vaadin-text-field>` already ships with themability, you could define pw-field with `...extends Polymer.GestureEventListeners(Vaadin.ThemableMixin(Vaadin.TextFieldElement))`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 8 of 16 files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-password-field.html, line 38 at r8](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn9Alc5ujiWFt8wKn_u:-Kn9d8mX4b5eidHu8eyj:bok9rj3) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/e3eee66d20895b49e448384e40e589c083fa098f/vaadin-password-field.html#L38)):*
> Eye password feature is implemented in all IE11 and Edge versions,

No.



---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 8 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-password-field.html, line 25 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8hmev6O-0egtC_JeQ:-Kn9ehfCQj390pNUSozs:b-xvuw73) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L25)):*

Ok, debugged this a little. The error in the current implementation is that VaadinTextField's template() function is never called when registering the pwf (the inheritance doesn't work as expected). Meaning that you should instead:

1. get text-field's template with `super.template.cloneNode(true)` and store it as the `memoizedTemplate` (you must name the pwf dom-module as something else that "vaadin-password-field" or else `super.template` will fetch that instead)
1. Get pwf template with Polymer.DomModule.import
1. Append the the "toggle-button" and `<style>`from pwf template to the `memoizedTemplate`
</blockquote></details>

Since you already wrote code and tested it, could you please submit it into by branch. I can't make it work :(

---

*[vaadin-password-field.html, line 71 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8i5I0gwUnZKuN4-An:-Kn9eyj3MavdilaMaxQ2:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L71)):*

Meaning that instead of relying that `<vaadin-text-field>` already ships with themability, you could define pw-field with `...extends Polymer.GestureEventListeners(Vaadin.ThemableMixin(Vaadin.TextFieldElement))`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

tomivirkki:
*[vaadin-password-field.html, line 25 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8hmev6O-0egtC_JeQ:-Kn9ifn1NVYCpsrDzcNg:b-xdyruy) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L25)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

Since you already wrote code and tested it, could you please submit it into by branch. I can't make it work :(
</blockquote></details>

Pushed the commit. Also make sure you have latest VaadinThemableMixin installed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
> I don't think replacing the native implementation with synthetic one is the best option here

It is the only one option here, since we want the component to be themable.

---

Review status: 8 of 16 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-Kn9if-fNxe3VPzPgJkP:b8bixs5)*

limonte:
Review status: 8 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-password-field.html, line 38 at r8](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn9Alc5ujiWFt8wKn_u:-Kn9isXRnToej9iAmiUE:byd461h) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/e3eee66d20895b49e448384e40e589c083fa098f/vaadin-password-field.html#L38)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

> Eye password feature is implemented in all IE11 and Edge versions,

No.


</blockquote></details>

Once again, in some versions there's NO native eye. If you don't believe me, I can demonstrate it to you via sharing screen session. Since we can't detect if native eye is implemented or not, removing custom eye for those users will disallow 'view password' feature at all.

Also, we want themable component with themable eye. Default one is not themable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 8 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-password-field.html, line 25 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8hmev6O-0egtC_JeQ:-Kn9jtwZZXaQgo0UvzN2:b-hnbqt4) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L25)):*

Pushed the commit. Also make sure you have latest VaadinThemableMixin installed.
</blockquote></details>

Many thanks, will update deps and try this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 8 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-password-field.html, line 25 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8hmev6O-0egtC_JeQ:-Kn9m-zQF_2x6KjmrDJ-:bmf7yzn) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L25)):*
> Also make sure you have latest VaadinThemableMixin installed.

Pulled your commit, reinstalled bower deps and it doesn't work for me. Eye is always hidden for `demo/password.html`. Does it work for you? 

Just to mention, not a blocker: inheritance of themable components doesn't look like a simple thing. Actually, it looks very tricky and hard to support/debug :(

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

tomivirkki:
*[vaadin-password-field.html, line 25 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8hmev6O-0egtC_JeQ:-Kn9nw6duGP7bLhZovGQ:bqd8pne) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L25)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

> Also make sure you have latest VaadinThemableMixin installed.

Pulled your commit, reinstalled bower deps and it doesn't work for me. Eye is always hidden for `demo/password.html`. Does it work for you? 

Just to mention, not a blocker: inheritance of themable components doesn't look like a simple thing. Actually, it looks very tricky and hard to support/debug :(
</blockquote></details>

Inheritance of themable components is as simple as `MyComponent extends VaadinTextField`. What makes it more complex is if you want to modify the superclass template but that's the case in any Polymer element inheritance, not just themable components. On top of that we add our own themable parts to the element which complicates it more but that doesn't relate to inheriting any way, just a standard practice of extending VaadinThemableMixin.

I re-cloned the repo and installed deps, works for me just fine. Which version of themable mixin are you running?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: 8 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-password-field.html, line 25 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8hmev6O-0egtC_JeQ:-Kn9rOj7pN-28lsx0Ozw:b-yo18q2) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L25)):*

Inheritance of themable components is as simple as `MyComponent extends VaadinTextField`. What makes it more complex is if you want to modify the superclass template but that's the case in any Polymer element inheritance, not just themable components. On top of that we add our own themable parts to the element which complicates it more but that doesn't relate to inheriting any way, just a standard practice of extending VaadinThemableMixin.

I re-cloned the repo and installed deps, works for me just fine. Which version of themable mixin are you running?

</blockquote></details>

Weird, works for me after re-cloning. Anyway, thanks.

PS. Don't forget to accept CLA.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

tomivirkki:
Reviewed 8 of 17 files at r2, 5 of 8 files at r5, 2 of 3 files at r8, 1 of 1 files at r10.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-password-field.html, line 25 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-Kn8hmev6O-0egtC_JeQ:-Kn9sBru69I_Yxu2avda:bl1h72v) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/71c9b0abf735ae22b9848ffba5d327f600e0cd56/vaadin-password-field.html#L25)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

Weird, works for me after re-cloning. Anyway, thanks.

PS. Don't forget to accept CLA.
</blockquote></details>

Ok, done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

manolo:
Ok, lets go with the synthetic implementation until we figure out a reliable way to useThePlatform. :lgtm: 

---

Reviewed 5 of 8 files at r5, 2 of 3 files at r8, 1 of 1 files at r10.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-KnDKFVt56Ky0Nn-X1uU:b-58ndlx)*

tomivirkki:
*[vaadin-password-field.html, line 95 at r10](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-KnE0BHt5R54iNBLxFek:-KnE0BHt5R54iNBLxFel:b8ob17q) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/fddd39883d5c891b4ae7902b85bf27cc74baa70b/vaadin-password-field.html#L95)):*
> ```HTML
>              * Set to true to hide the eye icon which toggles the password visibility.
>              */
>             hideToggleButton: {
> ```

Is this feature needed since you can hide the icon with CSS (requires you to add a "theme-for" module though)? @jouni? If we decide to keep the feature, it needs a wct-test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

tomivirkki:
*[vaadin-password-field.html, line 95 at r10](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-KnE0BHt5R54iNBLxFek:-KnE0wYCWxm0pXTGxrpK:b-epltp2) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/fddd39883d5c891b4ae7902b85bf27cc74baa70b/vaadin-password-field.html#L95)):*

Is this feature needed since you can hide the icon with CSS (requires you to add a "theme-for" module though)? @jouni? If we decide to keep the feature, it needs a wct-test.
</blockquote></details>

The default value should probably be false for this also

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
> Toggling the visibility feels flickery since the outline disappears while mouse button is pressed. Can something be done to fix this? I noticed it's a common problem with all prefixes/suffixes however.

https://github.com/vaadin/vaadin-control-state-mixin/issues/5

> The icon is quite hard to hit with mobile device (combo-box and date-picker icons assumably have the same issue).

This is actually related to the text-field. I will report the issue with description why this happening right after merging this PR.

---

Review status: 14 of 16 files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-password-field.html, line 95 at r10](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-KnE0BHt5R54iNBLxFek:-KnE6dPUP18is16eTdIj:b-b3qw8w) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/fddd39883d5c891b4ae7902b85bf27cc74baa70b/vaadin-password-field.html#L95)):*
> Is this feature needed since you can hide the icon with CSS

This was suggested by Manolo, I agree with him. It's good to have the ability to hide it without diving into themability and CSS.

>  If we decide to keep the feature, it needs a wct-test.

Added.

> The default value should probably be false for this also

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-KnDkRPojABuezEawVG6:blybbar)*

tomivirkki:
:lgtm:

---

Reviewed 2 of 2 files at r11.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-KnE7c1w1meVOA6xj6o8:bnfp4nl)*

Saulis:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[vaadin-password-field.html, line 54 at r11](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-KnYI6Mcq0o7tnflewcc:-KnYI6Mcq0o7tnflewcd:blyj9yx) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/1e07a23443854bcb92786264ca7a2eedcfcaafda/vaadin-password-field.html#L54)):*
> ```HTML
>     <div
>       part="toggle-button"
>       on-tap="_togglePasswordVisibility"
> ```

I would change this to `on-down` and then call `e.preventDefault()` on the handler to prevent the input field from blurring when clicked.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-password-field.html, line 54 at r11](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-KnYI6Mcq0o7tnflewcc:-KnYIoxTdz0xVWvQdZl1:b-6yo45e) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/1e07a23443854bcb92786264ca7a2eedcfcaafda/vaadin-password-field.html#L54)):*

I would change this to `on-down` and then call `e.preventDefault()` on the handler to prevent the input field from blurring when clicked.
</blockquote></details>

Can't agree, the action should not happen right after the `down` event. The issue with blurring should be handled in control-state-mixin because it happens not only for eye-icon, but for prefix and suffix as well. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

Saulis:
:lgtm: Let's fix the flickering issue in control state mixin.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-password-field.html, line 54 at r11](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-KnYI6Mcq0o7tnflewcc:-KnYQLm4ycpLXjKbIuFJ:b-gded2w) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/1e07a23443854bcb92786264ca7a2eedcfcaafda/vaadin-password-field.html#L54)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

Can't agree, the action should not happen right after the `down` event. The issue with blurring should be handled in control-state-mixin because it happens not only for eye-icon, but for prefix and suffix as well. 
</blockquote></details>

Discussed with @limonte, looks like we can remove extra `blur` and `focus` listeners from the `focusElement` now that `ShadyDOM` polyfill has been fixed to properly bubble `blur` and `focus` events from inside the shadow root. So no need to do changes here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-:-KnYQaQ_8OT4OQJH4Pjo:b-e4oxdt)*

Saulis:
Reviewed 8 of 17 files at r2, 5 of 8 files at r5, 1 of 3 files at r8, 2 of 2 files at r11.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

limonte:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-password-field.html, line 54 at r11](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98#-KnYI6Mcq0o7tnflewcc:-KnYS6xzqzO2RkiRqP0h:b-wn56uv) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/1e07a23443854bcb92786264ca7a2eedcfcaafda/vaadin-password-field.html#L54)):*

Discussed with @limonte, looks like we can remove extra `blur` and `focus` listeners from the `focusElement` now that `ShadyDOM` polyfill has been fixed to properly bubble `blur` and `focus` events from inside the shadow root. So no need to do changes here.
</blockquote></details>

https://github.com/vaadin/vaadin-control-state-mixin/pull/6

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/98)*

PR #110, opened 2017-08-08
vaadin-number-field element

gatanaso:
Connects to #296

This element is  a specification over the vaadin-text-field element that allows numeric value input. The vaadin-number-field element can be used as a simple input element with decrease/increase buttons to update the value. Additionally, it supports specifying min/max input value limits. For more details, please refer to the documentation and demo pages of the component.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/110)


demo/number-field.html
```diff
@@ -0,0 +1,42 @@
+<!doctype html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1" />
```
limonte: Issues like this should be fixed in skeleton and then applied w/o CR to all components we have: https://github.com/vaadin/vaadin-element-skeleton/blob/master/demo/advanced.html

test/visual/number-field.html
```diff
@@ -0,0 +1,25 @@
+<!doctype html>
```
limonte: Issues like this should be fixed in skeleton and then applied w/o CR to all components we have: https://github.com/vaadin/vaadin-element-skeleton/blob/master/demo/advanced.html

vaadin-number-field.html
```diff
@@ -0,0 +1,192 @@
...
+      }
+
+      [part="value"] {
+        -moz-appearance: textfield;
+      }
+    </style>
+
+    <div 
+      part="decrease-button" 
+      on-tap="_decreaseValue"
```
limonte: Replace `on-tap` with `on-click` and remove the `gesture-event-listeners.html` dependency`

The same as https://github.com/vaadin/vaadin-text-field/pull/108/files

tomivirkki: Why change this? Switching from tap to click makes clicking the buttons laggy on iOS

limonte: > Switching from tap to click makes clicking the buttons laggy on iOS

Are you 100% positive about that?

There's nothing about `on-tap` in [docs](https://www.polymer-project.org/2.0/docs/devguide/events) and I wasn't able to experience any lags on iOS with `on-click`

tomivirkki: Just tried with ios Simulator and the (300ms) lag seems to be there (with tap changed to click). Can someone verify?

abdonrd: Some references about this: https://github.com/Polymer/docs/pull/2283

tomivirkki: Ok, nice so the `tap` is being deprecated 👍  The [docs](https://www.polymer-project.org/2.0/docs/devguide/gesture-events) however still has it and the delay is still there on my simulator (iOS 10.3). In which IOS version should the delay disappear?

gatanaso: @limonte Updated on-tap to on-click and removed dependent library. Thanks for pointing this out!

demo/number-field.html
gatanaso: removed the extra "/" in the meta tag, thanks @samiheikki and @limonte

test/visual/number-field.html
gatanaso: capitalized doctype, thanks @samiheikki and @limonte

vaadin-number-field.html
abdonrd: @tomivirkki sorry, but I don't know. 😕

gatanaso:
@abdonrd Thank you very much for the suggestion. Latest commit fixes that issue.

abdonrd:
There are any ETA to publish this element?

jouni:
> There are any ETA to publish this element?

Not yet, sorry. But I think we will merge and tag a version within the next month. It might not be stable at that point, but hopefully beta quality at least.

vaadin-number-field.html
```diff
@@ -0,0 +1,199 @@
...
+        opacity: 0.5;
+      }
+
+      [part="decrease-button"][hidden],
+      [part="increase-button"][hidden] {
+        display: none;
+      }
+
+      [part="decrease-button"]::before {
+        content: "–"; /* en dash */
```
jouni: Use the proper minus symbol instead: −

For more info, if you’re a typogeek :)
http://www.punctuationmatters.com/the-difference-between-a-dash-and-a-minus-sign/

limonte: Done, thanks!

gatanaso:
*[vaadin-number-field.html, line 40 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110#-Kr0sCglwAZfzKPrs6Ox-r6-40:-KsgRtP8dJmW9_A0FluY:b-heqfhf) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a76295d89f042c4cbc2313bd2e12f340cc1be681/vaadin-number-field.html#L40)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

Done, thanks!
</blockquote></details>

Thanks for the info @jouni :)
And thanks @limonte for making the update!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110)*

manolo:
when clicking on control buttons, the component is focused (out line is shown) but not the native input, so you cannot write or use arrow keys.

---

Review status: 0 of 12 files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---

*[vaadin-number-field.html, line 97 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110#-KtWauUcGTTKaWTZlm-s:-KtWauUcGTTKaWTZlm-t:brt3gom) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/f9cc0fded6abcc98eb257025a8a4edbe5593a9bb/vaadin-number-field.html#L97)):*
> ```HTML
>           return {
>             /**
>             * Set to true to display value increase/decrease controls.
> ```

nit, jsdoc needs that you ident one space the second line and following ones

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110#-:-KtWc8qFzXpDESHsHDwF:b-2ji7n3)*

manolo:
Needs rebasing and pass CI

---

Review status: 0 of 12 files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110#-:-KtWisE5vNYs2sEP60vd:biqk1bp)*

vaadin-number-field.html
```diff
@@ -0,0 +1,199 @@
...
+              type: Number,
+              reflectToAttribue: true,
+              observer: '_maxChanged'
+            }
+          };
+        }
+
+        ready() {
+          super.ready();
+          this.$.input.type = 'number';
```
abdonrd: `this.focusElement.type = 'number';`

limonte: Done, thanks for noticing this! I'll try to break through our bureaucracy and get `vaadin-number-field` merged even if it's not in our current roadmap.

limonte:
> Needs rebasing and pass CI

Done.

---

Reviewed 2 of 10 files at r1, 17 of 18 files at r8.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-number-field.html, line 97 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110#-KtWauUcGTTKaWTZlm-s:-KuyEBAT5VRcyAp8ooGa:bb7f2mf) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/f9cc0fded6abcc98eb257025a8a4edbe5593a9bb/vaadin-number-field.html#L97)):*

nit, jsdoc needs that you ident one space the second line and following ones
</blockquote></details>

done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110#-:-KuyDhPkMGAw2lW_c0aT:b-w8w5wr)*

vaadin-number-field.html
abdonrd: Your welcome! :)

Sounds good!

abdonrd: @limonte offtopic question: Why Reviewable instead of GitHub reviews?

limonte: Reviewable works better for our internal processes. For external contributions, we are trying to use GitHub comments.

abdonrd: Ok, thanks!

abdonrd:
Any news? 🙄

Thanks in advance! 👏

swapnilbhattad:
I tried using this component, but its not accepting decimal numbers, tried to understand where it is validating integers, but couldn't understand.
Can you please help me in changing the component to allow decimal values as well.

abdonrd:
Any news? 🙄

jouni:
Sigh, it seems I was a bit hasty with my previous comment, that we would have time to merge this into the next version 😒

Should eventually be added, though. @tomivirkki, I think this is your call when you think this is ready to be merged.

The only open question for me is that should we support two separate designs, maybe even have it adapt responsively to the input device (touch vs mouse). Also, resolution/step is something that we probably want to add eventually (not necessarily in the first version).

screen shot 2018-03-12 at 10 17 29

sudipta1411:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110#-L7hFx3sPowuzFGUtbG3:-L7hFx3sPowuzFGUtbG4:b-1mzfdz) (no related file):*
If the min property is set to 0, then one is able to decrease past the min value. This is because the check `if(this.min)` does not pass.

---

*[vaadin-number-field.html, line 99 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110#-L7dDfd918Gjj4gs5p-J:-L7dDfd918Gjj4gs5p-K:bex4ena) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/a76295d89f042c4cbc2313bd2e12f340cc1be681/vaadin-number-field.html#L99)):*
> /
>             h

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110)*

manolo:
Updated PR  so as it fits latest element structure, and we can continue the discussion, no new features added, apart from  css moved to the corresponding theme file

---

Review status: 1 of 15 files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-number-field.html, line 65 at r8](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110#-L7s79Iha8Mo78SWmxEM:-L7s79Iha8Mo78SWmxEN:brducp) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/d99754152608b7e73996b480f6749b68b775e1f4/vaadin-number-field.html#L65)):*
> ```HTML
>     </div>
> 
>     <div
> ```

why not use `slot` someway for buttons so as they are automatically inserted in the correct  position and themed appropriately ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110#-:-L7sZBOEKbeOkhvG59WD:b737f3s)*

manolo:
Review status: 0 of 15 files reviewed at latest revision, 8 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110#-L7hFx3sPowuzFGUtbG3:-L7xeB3dAvkmFwBxzJL1:bgtvwd1) (no related file):*
<details><summary><i>Previously, sudipta1411 (Sudipt Roy) wrote…</i></summary><blockquote>

If the min property is set to 0, then one is able to decrease past the min value. This is because the check `if(this.min)` does not pass.
</blockquote></details>

@sudipta1411 fixed the issue, thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/110)*

jouni:
2) Disable the corresponding button when a limit is reached

3) Should probably allow left/right keys as well, as the buttons visually indicate those directions

manolo:
> Disable the corresponding button when a limit is reached

That is already implemented this PR take a look to last commit. Since it's a div there is a disabled attribute that takes an opacity of 0.2 in theme, it can be changed though.

> When value is same is min or max, clicking the controls doesn't give any feedback to user

Now, it does, by changing the style of the button. You might to update your local bower_components with last branch changes

> Should probably allow left/right keys as well, as the buttons visually indicate those directions

up/down are handled natively by `input type=number` so no need of event listeners, etc, I think adding key-listeners for non the non-native behavior that needs some extra code and tests,  probably isn't worth 

> Updating the value through controls does not trigger change event on element.

Will fix that in this PR

> Support for step attribute.

I would incorporate that in separated PR after some discussion about API and UX design of that feature

> Stop user from entering value outside min and max range.

That can be handled with the regexp pattern 

---

Review status: 0 of 15 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/110#-:-LC45PUS3zGrefOmbv6l:be3luli)*

manolo:
- Rebased with master `2.3.0-alpha1` 
- Fixed focus on click buttons
- Fixed change event on click buttons
- Added step feature

test/number-field.html
```diff
@@ -0,0 +1,254 @@
...
+          const changeSpy = sinon.spy();
+          numberField.addEventListener('change', changeSpy);
+
+          increaseButton.click();
+          expect(changeSpy.callCount).to.equal(1);
+        });
+
+        it('should not focus input when a button is clicked', () => {
+          let hasFocus = false;
+          input.focus = () => {
```
web-padawan: nit: would be better to use `sinon.spy(input, 'focus')` instead

theme/lumo/vaadin-number-field-styles.html
```diff
@@ -0,0 +1,29 @@
...
+
+<dom-module id="lumo-number-field" theme-for="vaadin-number-field">
+  <template>
+    <style include="lumo-field-button">
+      :host {
+        width: 8em;
+      }
+
+      :host([has-controls]) [part="value"] {
+        text-align: center;
```
web-padawan: note: it makes sense to include most of these styles below (excluding font-size and padding) into the core styles, as they are currently copy-pasted in both Lumo and Material

PR #121, opened 2017-08-24
Mask the password on blur

### LINKED ISSUE #116: [password-field] Mask the password on blur
When the user blurs `<vaadin-password-field>` the input should be masked again, to avoid leaving the password visible accidentally.

For this to work nicely, the field needs to be focused when the reveal button is pressed, like it currently is, but that could change in a future update/fix to vaadin-text-field. So it could make sense to add a test for it.

Can anyone think of a use case where this would be a poor default? And if yes, should this behavior be configurable?

limonte:
Fixes #116

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/121)

### DIFF
--- a/test/password-field.html
+++ b/test/password-field.html
@@ -46,7 +46,7 @@
           toggleButton.click();
           expect(input.type).to.equal('text');
 
-          passwordField.blur();
+          passwordField.dispatchEvent(new Event('blur'));
           expect(input.type).to.equal('password');
         });
       });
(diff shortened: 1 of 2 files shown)


tomivirkki:
:lgtm_strong: 

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/121#-:-KsNZuD0Ajly8UIa1j1Q:b-uwkowd)*

PR #131, opened 2017-09-05
Add ability to prevent focus on prefix/suffix click

limonte:
This PR is a dependency for https://github.com/vaadin/vaadin-combo-box/pull/507

`control-state-mixin` listens to the `click` event on host and focus the `focusElement` after click. This PR will stop `click` event propagation from prefix/suffix to the host if `prevent-focus` attribute is set.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/131)

### DIFF
--- a/vaadin-text-field.html
+++ b/vaadin-text-field.html
@@ -92,7 +92,7 @@ This program is available under Apache License Version 2.0, available at https:/
 
     </style>
 
-    <div class="container">
+    <div class="container" id="container">
 
       <label part="label" on-click="focus" id="[[_labelId]]">[[label]]</label>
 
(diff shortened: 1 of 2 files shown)


vaadin-text-field.html
```diff
@@ -347,6 +347,12 @@
           this._setHasValue(newVal !== '' && newVal != null);
         }
 
+        _clickHandler(e) {
+          if (e.target.hasAttribute('prevent-focus')) {
```
jouni: I think this won’t catch cases when there are nested elements inside the prefix/suffix element, for example: 

```html
<vaadin-text-field>
  <vaadin-button prevent-focus>
    <iron-icon icon="plus"></iron-icon>
  </vaadin-button>
</vaadin-text-field>
```

The target would most probably be the `iron-icon` without the attribute.

limonte:
This isn't a good approach, closing.

PR #149, opened 2017-09-27
Fallback to invalid property in checkValidity() if no constraints are…

### LINKED ISSUE #130: Do not validate vaadin-text-field (and vaadin-password-field) when no pattern and required attributes are present.
I would like to be able to set the `invalid` status manually, for instance, this way:

`<vaadin-text-field id="textField" value={{text}}></vaadin-text-field>`

```js
// Omitted most of the file for clarity

static get properties() {
    return {
        text: {
            type: String,
            observer: '_textChangeObserver'
        }
    }
}

_textChangeObserver(newValue, oldValue) {
    if (newValue && newValue.startsWith('2')) {
        this.$.textField.invalid = true;
        this.$.textField.errorMessage = "Input should not start with 2";
    } else {
        this.$.textField.invalid = false;
        this.$.textField.errorMessage = "";
    }
}
```

but this is not possible, since every time the component looses focus, it revalidates itself.

Since there were no attributes on the component that may change the validation result, I would have expected the component not to carry out any validation at all.

Can you please fix this?

tomivirkki:
Fixes #130

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/149)

### DIFF
--- a/test/validation.html
+++ b/test/validation.html
@@ -76,6 +76,28 @@
         expect(tf.validate()).to.be.true;
       });
 
+      it('should check validity on validate', function() {
+        const spy = sinon.spy(tf, 'checkValidity');
+        tf.validate();
+        expect(spy.called).to.be.true;
+      });
+
+      it('should not change invalid property is no constraints are set', function() {
+        tf.validate();
+        expect(tf.invalid).to.be.false;
+        tf.invalid = true;
+        tf.validate();
+        expect(tf.invalid).to.be.true;
+      });
+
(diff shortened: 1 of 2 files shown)


limonte:
:lgtm_strong: 

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/149#-:-Kv1N62srpZe-Y4BaZ1Q:b-uwkowd)*

PR #174, opened 2017-11-28
Make sure the field has focus when revealed

### LINKED ISSUE #116: [password-field] Mask the password on blur

tomivirkki:
Fixes #116

There was a regression in v1.1.0 that caused the field not getting focused on password reveal (if not focused beforehand). The issue is described in [this comment](https://github.com/vaadin/vaadin-text-field/issues/116#issuecomment-346966907)

This PR fixes the regression

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/174)

### DIFF
--- a/test/password-field.html
+++ b/test/password-field.html
@@ -59,10 +59,17 @@
           expect(revealButton.hidden).to.be.true;
         });
 
-        it('should prevent mousedown event on reveal-button', function() {
+        it('should prevent mousedown event on reveal-button when focused', function() {
           const e = new CustomEvent('mousedown', {bubbles: true, cancelable: true});
+          passwordField.focus();
           revealButton.dispatchEvent(e);
           expect(e.defaultPrevented).to.be.true;
+        });
+
+        it('should not prevent mousedown event on reveal-button when not focused', function() {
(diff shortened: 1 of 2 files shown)


web-padawan:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/174#-:-L-BmrIqlDZk7eIo7UNr:bnfp4nl)*

vicsstar:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/174#-:-L-C8_7lK4QKCPW32A3I:bnfp4nl)*

PR #176, opened 2017-12-14
Ensure text area styles override text field styles in text area

tomivirkki:
The order in which matching style modules are included in an element template is not consistent.

So the current approach makes the theme look broken in case text field styles get included _after_ text area specific styles (that are supposed to override text-field styles) in the text area template.

Need to add an explicit include for text field valo styles prior to text-area specific styles.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/176)

### DIFF
--- a/theme/valo/vaadin-text-area.html
+++ b/theme/valo/vaadin-text-area.html
@@ -5,7 +5,7 @@
 
 <dom-module id="valo-text-area" theme-for="vaadin-text-area">
   <template>
-    <style>
+    <style include="valo-text-field">
       :host {
         width: calc(var(--valo-size-m) * 10);
       }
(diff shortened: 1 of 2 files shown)


jouni:
This is actually a much better approach in general, that reusable modules are included explicitly by the actual element theme, and not “forced” by the reusable module.

web-padawan:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/176#-:-L0JbBdIbuxZq7UkKlN7:bnfp4nl)*

YuriyVaadin:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/176#-:-L0KCx43aa_YWosDuFPS:bnfp4nl)*

PR #177, opened 2017-12-18
Update theme

jouni:
- Extract shared styles to a style module in to vaadin-text-field-mixin

- Reset the native value element styles so it appears as a div/span for a theme developer

- Make vertical alignment work as expected for all included elements. They now sit nicely on the baseline in all browsers. Minor mis-alignment with vaadin-button in IE11, which can be looked at later.

- Updated visual look’n’feel (FFS), remove all box-shadows (excluding focus-ring)

- Remove “contrast” color variation (still in the demo, will be removed in an upcoming update to the demos)

- Use the `—valo-icon-size` property for sizing the icons (will be in the next Valo version)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/177)


tomivirkki:
:lgtm_strong: 

---

Reviewed 12 of 12 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/177#-:-L0dz0u76I_OQ5of21xs:b-uwkowd)*

PR #188, opened 2018-01-04
Fix text area activation effect

jouni:
Since the pseudo element doesn’t scroll with the input field, we’ll just use a background-color transition for the hover/active effect.

The text-field transition can actually look too distracting on large text areas, so this is probably even a better design.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/188)


web-padawan:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/188#-:-L24WQZsa7mlHXVCzduv:bnfp4nl)*

PR #194, opened 2018-01-17
Add has-label attribute for styling

### LINKED ISSUE #186: Reflect label as a state attribute
- Add a test
- Add to documentation
- Update analysis.json

This fix is need by some Lumo selectors. For example: https://github.com/vaadin/vaadin-text-field/blob/master/theme/lumo/vaadin-text-field.html#L51

web-padawan:
Fixes #186

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/194)

### DIFF
--- a/analysis.json
+++ b/analysis.json
@@ -8,20 +8,106 @@
       "sourceRange": {
         "file": "src/vaadin-text-area.html",
         "start": {
-          "line": 178,
+          "line": 179,
           "column": 6
         },
         "end": {
-          "line": 178,
+          "line": 179,
           "column": 42
         }
       },
       "elements": [
         {
-          "description": "`<vaadin-text-field>` is a Polymer 2 element for text field control in forms.\n\n```html\n<vaadin-text-field label=\"First Name\">\n</vaadin-text-field>\n```\n\n### Styling\n\nThe following custom properties are available for styling:\n\nCustom property | Description | Default\n----------------|-------------|-------------\n`--vaadin-text-field-default-width` | Set the default width of the input field | `12em`\n\nThe following shadow DOM parts are available for styling:\n\nPart name | Description\n----------------|----------------\n`label` | The label element\n`input-field` | The element that wraps prefix, value and suffix\n`value` | The text value element inside the `input-field` element\n`error-message` | The error message element\n\nThe following state attributes are available for styling:\n\nAttribute    | Description | Part name\n-------------|-------------|------------\n`disabled` | Set to a disabled text field | :host\n`has-value` | Set when the element has a value | :host\n`invalid` | Set when the element is invalid | :host\n`focused` | Set when the element is focused | :host\n`focus-ring` | Set when the element is keyboard focused | :host\n`readonly` | Set to a readonly text field | :host\n\nSee [ThemableMixin – how to apply styles for shadow parts](https://github.com/vaadin/vaadin-themable-mixin/wiki)",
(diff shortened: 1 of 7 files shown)


tomivirkki:
*[src/vaadin-text-field.html, line 93 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/194#-L37LftGlHEDpPC04fko:-L37LftGlHEDpPC04fkp:b-a2b411) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/e3e205861b5ffbb21dd912dca1b97df1c5c3f963/src/vaadin-text-field.html#L93)):*
> ```HTML
>        * `disabled` | Set to a disabled text field | :host
>        * `has-value` | Set when the element has a value | :host
>        * `label` | Set when the element has a label | :host
> ```

After reconsidering, would it make more sense to align the behaviour with that of `value` and just mark the element with `has-label`. The value of the `label` property is irrelevant for the theme.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/194)*

web-padawan:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/vaadin-text-field.html, line 93 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/194#-L37LftGlHEDpPC04fko:-L37XlwGghnT2KhrxXP0:b-896fix) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/e3e205861b5ffbb21dd912dca1b97df1c5c3f963/src/vaadin-text-field.html#L93)):*

After reconsidering, would it make more sense to align the behaviour with that of `value` and just mark the element with `has-label`. The value of the `label` property is irrelevant for the theme.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/194)*

tomivirkki:
:lgtm:

---

Reviewed 7 of 7 files at r1, 6 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/194#-:-L37Yl5veC3HdorWvOQw:bnfp4nl)*

platosha:
Now that `has-label` makes us to edit the theme anyway, would it rather make sense to simply hide the `<label>` when it’s text is empty?

If done this way, the theme can define spacing in a more straightforward way, without having to use selectors like `[label]`. No API changes required.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/194#-:-L37ZoulPh-5feFUJNij:bfwd0fu)*

web-padawan:
After discussing we came up to conclusion that adding `has-label` would make things complex from the a11y point of view when using in`vaadin-form-layout` with the external label.

@jouni what if we just set`position: absolute` on the label part and `align-items: baseline;` on the host element? It appears to be simpler than the current implementation with two nested flex containers and pseudo element pushing element without label to fit baseline. The drawback is that we would still need to add `margin-top` for element having label, and with the absolute positioning this is not possible.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/194#-:-L37chQzRq4WXR4XNr6l:b-23zo9d)*

jouni:
I’m not in favor of moving the limitation on the conditional margin (based on the existence of a label).

> Now that has-label makes us edit the theme anyway, would it rather make sense to simply hide the <label> when it’s text is empty?

This is what I actually wanted to do in Lumo in the beginning, but since the default value was previously `undefined` the `<label>` element was never empty even if no label was specified. Then I change the default value to `''` and made the dependent elements bind to the `label` attribute instead of the property.

But I learned that we can use `null` as the default value, and everything should work nicely after that. There’s already `[part="label"]:empty { display: none; }` in Lumo. I suppose we can then revert back to binding to the `label` property in the dependent elements, if wanted.

web-padawan:
So it appears that preserving the `margin` on the element depending on whether it has label or not would still require us to have this attribute, and the `vaadin-material-theme` relies on it as well for some transitions: https://github.com/vaadin/vaadin-material-theme/blob/master/vaadin-text-field.html#L111

platosha:
Right, explicit `[has-label]` is good to have then :thumbsup: 

:lgtm: 

---

Reviewed 1 of 7 files at r1, 6 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-text-field/194#-:-L3CTEV18n5AEnohHVC9:bkifxmv)*

Issue #212, opened 2018-03-30
Add support for a label "suffix" slot


heruan:
It would be useful to have support to add components to the label part of a text field to a text field in an unobtrusive manner, e.g. to add additional information or to provide field-related actions.

Use cases:

<img width="262" alt="screen shot 2018-03-30 at 19 33 40" src="https://user-images.githubusercontent.com/851745/38147153-4a01d5ce-3451-11e8-8b4f-2c281deb6886.png">

```html
<vaadin-text-field label="Search">
  <a slot="label-suffix" href="#">help</a>
</vaadin-text-field>

<vaadin-password-field label="New password" required>
  <a slot="label-suffix" href="#">suggest</a>
</vaadin-password-field>

<vaadin-combo-box label="Choose user" required>
  <a slot="label-suffix" href="#">me</a>
</vaadin-combo-box>
```

See: https://jsfiddle.net/heruan/nry6nu00/

jouni:
Is there possibly a more general use case here, about customizing the whole label, and not just the suffix?

At least I have considered that the current `<vaadin-text-field label="...">` attribute is just a convenience API for setting the text content of a label slot internally.

Perhaps the suffix use case is the most common one, but I can at least think of a use case that I want to provide an icon before the label text which seems fairly common as well.

So maybe it should be `slot="label"` instead?

```html
<vaadin-text-field>
  Search <a slot="label" href="#">help</a>
</vaadin-text-field>
```

This forces users to provide some additional CSS to align the link to the end, and possibly to handle the required indicator somehow as well. But I feel like we should consider this more generic use case.

heruan:
I agree with the more generic `label` slot, but I'd still keep the `label-suffix` slot as a convenient way to avoid this:

> This forces users to provide some additional CSS to align the link to the end, and possibly to handle the required indicator somehow as well.

What do you think?

jouni:
I’m not keen on having too many slots. I’m concerned about the complexity and performance implications it could mean. Also, are nested slots supported?

To align the suffix in my previous example, this should be enough:

```html
<dom-module id="text-field-label-suffix" theme-for="vaadin-text-field">
  <template>
    <style>
      [part="label"] {
        display: flex;
        justify-content: space-between;
      }
    </style>
  </template>
</dom-module>
```

For the required indicator, perhaps we have a special attribute you can add to an element to indicate that it should be used:

```html
<vaadin-text-field>
  <div slot="label">
[...]

PR #213, opened 2018-03-30
Add a "label-suffix" slot

heruan:
Implementation of #212.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/213)

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -36,8 +36,19 @@ This program is available under Apache License Version 2.0, available at https:/
         width: var(--vaadin-text-field-default-width, 12em);
       }
 
-      [part="label"]:empty {
+      :host(:not([has-label])) [part="label-container"] {
         display: none;
+      }
+
+      [part="label-container"] {
+        display: flex;
+        justify-content: flex-end;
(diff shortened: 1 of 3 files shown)


tomivirkki:
I would suggest you to write a custom extension for this feature instead. Including this feature in the core text-field element would require a lot of consideration regarding Lumo/Material themes and since it's a new feature and also could be considered a breaking change, this can't be included during the ongoing beta period.

tomivirkki:
A breaking change would in practice require a new major release so unfortunately it's unlikely to make it in `<vaadin-text-field>` 2.X.

heruan:
I tried to avoid breaking changes, in fact I don't understand entirely which part of the changes you consider would break existing code using the element. I'm wrapping the `label` part in a `div`, but `part="label"` is still targetable, so existing themes unaware of the wrapper shouldn't break.

heruan:
@tomivirkki any feedback about what may break merging this PR? I'm willing to try to rework this, if the feature is something you'd consider at least.

heruan:
FWIW, I just fixed the label size issue for `vaadin-text-area` in the latest commit (not showing up here since the PR is closed now).

PR #245, opened 2018-08-01
Use @vaadin/vaadin-component-dev-dependencies

limonte:
The same as https://github.com/vaadin/vaadin-element-skeleton/pull/156

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/245)


web-padawan:
Visual tests are failing on master also, ok to fix in a separate PR.

PR #252, opened 2018-08-25
Replace input's and text-area's aria-labelledby attribute with aria-label

samiheikki:
Fixes https://github.com/vaadin/vaadin-date-picker/issues/534

Element is failing a11y tests described in the original ticket. Disclaimer: I'm not expert on a11y related materials, but at least with this approach text-field tests and the a11y tests are both passing.

I did not add a11y tests to the test suite as I feel that should be done to vaadin-element-skeleton first (if added at all).

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/252)

### DIFF
--- a/src/vaadin-text-area.html
+++ b/src/vaadin-text-area.html
@@ -64,7 +64,7 @@ This program is available under Apache License Version 2.0, available at https:/
             on-input="_onInput"
             on-change="_onChange"
             aria-describedby$="[[_getActiveErrorId(invalid, errorMessage, _errorId)]]"
-            aria-labelledby$="[[_getActiveLabelId(label, _labelId)]]"
+            aria-label$="[[label]]"
             aria-invalid$="[[invalid]]"></textarea>
 
         <slot name="suffix"></slot>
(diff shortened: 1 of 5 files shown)


samiheikki:
@limonte Removed `_getActiveLabelId `

web-padawan:
I'm a bit confused with the statement here: https://www.w3.org/TR/wai-aria-1.1/#aria-label

> If the label text is visible on screen, authors SHOULD use aria-labelledby and SHOULD NOT use aria-label.

Will update the PR with the tests, however a real expert opinion is needed here. I could try to investigate why `accessibility-developer-tools` fail on this, most likely because of Shadow DOM.

web-padawan:
Actually, this check is false positive because of Shadow DOM unfriendliness, exactly as I supposed.
See the lint there `axs` uses `ownerDocument.getElementById()` to check the element:

https://github.com/GoogleChrome/accessibility-developer-tools/blob/3d7c96bf34b3146f40aeb2720e0927f221ad8725/dist/js/axs_testing.js#L1456

See also GoogleChrome/accessibility-developer-tools#272

web-padawan:
Confirmed by @limonte that the current approach works in NVDA, both for label and error message. Therefore, closing this PR per above comment regarding ARIA guidance on `aria-labelledby` usage.

PR #253, opened 2018-08-27
Remove will-change property to prevent grid scroll issue

web-padawan:
Fixes scroll handle not visible when used inside of the grid (Mac OS Chrome)

The same has been already applied in
vaadin/vaadin-button#83
vaadin/vaadin-checkbox#83

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/253)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ script:
           wct --env saucelabs;
         fi;
       else
-        wct;
+        xvfb-run -s '-screen 0 1024x768x24' wct;
       fi &&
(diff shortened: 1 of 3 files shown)


tomivirkki:
Does this relate to https://github.com/vaadin/vaadin-grid/issues/1238?

web-padawan:
Could be related but not necessarily. Note the PRs mentioned above have been merged after vaadin/vaadin-grid#1238 was submitted, and the grid in "kitchen sink" only uses checkboxes.

CRUD demo of the grid has this issue currently.

PR #258, opened 2018-09-19
Add text selection API

jouni:
Fixes #214 

No examples for these, as they are quite low-level APIs.

Didn’t know if we should add tests for these, as they mostly just forward the functionality to the native element.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/258)

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -380,7 +380,6 @@ This program is available under Apache License Version 2.0, available at https:/
      *                            `"none"` if the direction is unknown or irrelevant.
      */
     setSelectionRange(start, end, direction) {
-      this.focus();
       this.focusElement.setSelectionRange(start, end, direction);
     }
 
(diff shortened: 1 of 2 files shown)


tomivirkki:
Thanks for the PR

> Didn’t know if we should add tests for these, as they mostly just forward the functionality to the native element.

Even if the code just forwards the native functionality, we need to have tests to make sure the code is there + does what it's expected to do.

jouni:
@tomivirkki , I don’t really understand why the visual tests are failing. I can’t spot any obvious differences with my local rendering and the reference screenshots. The changes shouldn’t affect the rendering at all either.

jouni:
> Why is the method called `selectAll()` and not `select()`?

No particular reason. But we should align with the native APIs for sure.

web-padawan:
It has been a while since this PR was submitted. Apologies about keeping it open for so long without any progress. 

The component has been converted to Polymer 3 so there are merge conflicts. 
Also, the new development for Vaadin web components will happen in the [monorepo](https://github.com/vaadin/vaadin-web-components). 

I'm closing this now. We will submit a new PR to the monorepo when the linked issue is prioritised.

PR #248, opened 2018-08-19
Align text field to baseline

samiheikki:
Fixes #246

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/248)

PR #262, opened 2018-10-14
Align prefix and suffix to center of text-area

samiheikki:
Fixes https://github.com/vaadin/vaadin-text-field/issues/261

Requires changes from https://github.com/vaadin/vaadin-text-field/pull/248 and update visuals

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/262)


samiheikki:
Good point, thanks!

PR #266, opened 2018-10-30
Change event by user input timeout to support `LAZY value change mode` - WIP

kumm:
A concept with a simple demo.
See https://github.com/vaadin/vaadin-text-field-flow/issues/114


kumm:
IE11 input event bug with placeholder is not analyzed yet, but should affect this function.

PR #268, opened 2018-11-02
Add support for the `type` attribute

### LINKED ISSUE #204: Enable setting type of input field
Right now it's not possible to create a email type vaadin-text-field without some javascript
to find the field and then executing `foundFieldElement.shadowRoot.querySelector('input').type = 'email'`.

One could expect that having `<vaadin-text-field type='email'></vaadin-text-field>` would result in the input field being of type `email`

jouni:
Fixes #204 

Opening the discussion if we could add this support, even as simple as this (no checking or black/whitelisting of the supported type).

Should probably add a test for this, but I’m avoiding extra work for now.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/268)

### DIFF
--- a/src/vaadin-text-field.html
+++ b/src/vaadin-text-field.html
@@ -39,6 +39,7 @@ This program is available under Apache License Version 2.0, available at https:/
           aria-required$="[[required]]"
           value="{{value::input}}"
           title="[[title]]"
+          type="[[type]]"
           on-blur="validate"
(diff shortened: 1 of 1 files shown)


web-padawan:
Decided to go with the slotted input approach instead, prototype done in #274 
Let me close this PR and continue discussion in that issue.

Issue #269, opened 2018-11-02
Focusing an instance that hasn't yet been attached throws


tomivirkki:
The following throws an error because the instance hasn't yet been attached to the DOM / fully initialized
```
document.createElement('vaadin-text-field').focus()
```

PR #270, opened 2018-11-03
Don't throw an error when focusing a newly created text-field

samiheikki:
Connects to #269 

Requires first this to be merged and released: https://github.com/vaadin/vaadin-control-state-mixin/pull/47

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/270)

### DIFF
--- a/test/text-field.html
+++ b/test/text-field.html
@@ -226,5 +226,12 @@
         });
       });
     });
+
+    describe('methods', function() {
+      it('should not throw an error when using focus() to a newly created element', () => {
+        // No expect needed as an error is thrown when focusing undefined element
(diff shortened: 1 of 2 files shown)


samiheikki:
1: With this fix it will behave the same way as native input `document.createElement('input').focus()`
2: That is mentioned in PR description

test/text-field.html
```diff
@@ -226,5 +226,12 @@
         });
       });
     });
+
+    describe('methods', function() {
+      it('should not throw an error when using focus() to a newly created method', () => {
```
limonte: Did you mean "newly created element"?

PR #271, opened 2018-11-04
Dispatch `iron-resize` event on text-area height change

### LINKED ISSUE #267: Text area should notify host components of resize
### Description
`<vaadin-text-area>` inside `<vaadin-grid>`'s row details doesn't work as expected. When the text-area gets resized in effect of a value change, the details row isn't resized respectively. This results into the grid's rows being misaligned.

### Expected outcome
Auto-resizing `<vaadin-text-area>` should affect the grid's row positioning/sizing upon resize.

### Actual outcome
The rows get misaligned

details-text-area

### Solution Proposal
Dispatch bubbling `iron-resize` events whenever the `<vaadin-text-area>` size gets updated.

samiheikki:
Fixes #267

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/271)

### DIFF
--- a/src/vaadin-text-area.html
+++ b/src/vaadin-text-area.html
@@ -186,7 +186,23 @@ This program is available under Apache License Version 2.0, available at https:/
           input.style.removeProperty('max-width');
           inputField.style.removeProperty('display');
           inputField.scrollTop = scrollTop;
+
+          if (this.__previousInputHeight && this.__previousInputHeight !== inputHeight) {
+            this.dispatchEvent(
+              new CustomEvent('iron-resize', {
+                bubbles: true
+              })
+            );
+          }
+
+          this.__previousInputHeight = inputHeight;
         }
+
+        /**
+         * Fired when the text-area height changes.
+         *
+         * @event iron-resize
+         */
       }
 
       customElements.define(TextAreaElement.is, TextAreaElement);
--- a/test/text-area.html
+++ b/test/text-area.html
@@ -291,5 +291,36 @@
         ));
       });
     });
+
+    describe('resize', () => {
+      let textArea, spy;
+
+      beforeEach(() => {
+        textArea = fixture('default');
+        spy = sinon.spy();
+        textArea.addEventListener('iron-resize', spy);
+      });
+
+      it('should not dispatch `iron-resize` event on init', () => {
+        expect(spy.callCount).to.equal(0);
+      });
+
+      it('should dispatch `iron-resize` event on height change', () => {
+        textArea.value = `
+        there
+        should
+        be
+        a
+        lot
+        of
+        rows`;
+        expect(spy.callCount).to.equal(1);
+      });
+
+      it('should not dispatch `iron-resize` event on value change if height did not change', () => {
+        textArea.value = 'just 1 row';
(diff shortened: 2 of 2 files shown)


tomivirkki:
___
*[src/vaadin-text-area.html, line 190 at r1](https://reviewable.io/reviews/vaadin/vaadin-text-field/271#-LQY2XnD6nFVy3_UOY8S:-LQY2XnE-jD7JafrHJIQ:bw7hs7e) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/470bdd5a3e65a88ecf3d9afc81cebde99f62317a/src/vaadin-text-area.html#L190)):*
> ```HTML
>           inputField.scrollTop = scrollTop;
> 
>           if (this.__previousInputHeight && this.__previousInputHeight !== inputHeight) {
> ```

Please test all new code. There should be additional tests to validate the checks in this if clause:
- "should not dispatch an iron-resize event on value change if height didn't change"?
- "should not dispatch an iron-resize event on init"?

tomivirkki:
___
*[src/vaadin-text-area.html, line 190 at r1](https://reviewable.io/reviews/vaadin/vaadin-text-field/271#-LQY2XnD6nFVy3_UOY8S:-LQhFKk264Oi99k6nS6N:bfqfk8a) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/470bdd5a3e65a88ecf3d9afc81cebde99f62317a/src/vaadin-text-area.html#L190)):*
<details><summary><i>Previously, samiheikki (Sami Suo-Heikki) wrote…</i></summary><blockquote>

Good point. Added missing tests
</blockquote></details>

The first `this.__previousInputHeight` still isn't validated by any test. The whole test suite passes just fine without the check

tomivirkki:
___
*[src/vaadin-text-area.html, line 193 at r3](https://reviewable.io/reviews/vaadin/vaadin-text-field/271#-LQhvA3r3KBfiMws2NzY:-LQi0d0T8gF7YEtU82OE:b-fudngv) ([raw file](https://github.com/vaadin/vaadin-text-field/blob/c232d394fe20ddca516f0c02e1cf84dbaed10852/src/vaadin-text-area.html#L193)):*
<details><summary><i>Previously, web-padawan (Serhii Kulykov) wrote…</i></summary><blockquote>

in the `IronResizableBehavior` this element has `bubbles: false` and `composed: true`

See the line where it happens: https://github.com/PolymerElements/iron-resizable-behavior/blob/7cef76323dcb2dd85cb72f40e578071e6e3edfe0/iron-resizable-behavior.html#L181

Note: the `composed` flag is set by using `fire` method, see this line: https://github.com/Polymer/polymer/blob/8882c3db248634403453e2969d1e9e02f5333b48/lib/legacy/legacy-element-mixin.html#L403

We will most likely have to add tests with the text-area wrapped in the other component to ensure the event retargeting works.
</blockquote></details>

If we won't bubble the `iron-resize` however, the implementation gets more complicated as we'll have to add more logic to grid to assign parent resizables for any content that gets (asynchronously) rendered in the grid cells. The suggested approach is kind of a lightweight workaround for the initial problem.

PR #277, opened 2018-11-13
Add clear button

### LINKED ISSUE #92: Built-in “clear” button
As an opt-in feature (or opt-out, can be discussed), provide a built-in suffix element which, when clicked/tapped, sets the value of the field to an empty string.

The element should be an “X” icon and be shown before any other suffix elements (closest to the value).

Question: When the field is marked as required, should the clear button always be hidden?

yuriy-fix:
Fixes #92

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/277)


demo/text-area-demos.html
```diff
@@ -48,6 +48,18 @@ <h3>Minimum Height</h3>
       </template>
     </vaadin-demo-snippet>
 
+    <h3>Display the Clear Button</h3>
```
jouni: I’d think this is a redundant example, as the top of the page says that text area has all the same features as text field. For example, we are not duplicating examples for label, required, or any other attribute.

Same thing for password field examples.

limonte: > Same thing for password field examples.

Agree about the textarea, but I was personally interested in the password field example, particularly how the clear button and the reveal-password button look and work together.

src/vaadin-text-area.html
```diff
@@ -67,6 +67,7 @@
             aria-labelledby$="[[_getActiveLabelId(label, _labelId)]]"
             aria-invalid$="[[invalid]]"></textarea>
 
+        <div part="clear-button" slot="suffix" id="clearButton" role="button" aria-label="Clear" hidden$=[[!clearButtonVisible]]></div>
```
tomivirkki: `slot` not needed

src/vaadin-text-field-mixin.html
```diff
@@ -323,6 +348,17 @@
       return !(this.invalid = !this.checkValidity());
     }
 
+    clear(e) {
```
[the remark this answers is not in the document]
yuriy-fix: Updated. No parameters anymore.

src/vaadin-text-field-mixin.html
```diff
@@ -303,6 +325,9 @@
         this.updateStyles();
       }
 
+      this.$.clearButton.addEventListener('mousedown', this.clear.bind(this));
```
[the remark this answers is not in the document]
yuriy-fix: Was done in order to workaround `change` event on host. Switched to `click` now and aligned with `combo-box`

demo/text-area-demos.html
yuriy-fix: Removed from password-field and text-area demos, but can restore at any point of time

src/vaadin-text-area.html
yuriy-fix: Removed.

Issue #281, opened 2018-11-22
Implement slotted input feature


web-padawan:
Child of #273

The slotted input should work with both text-field and password-field. Slotted textarea should work with text-area.

1. Allow user to slot the `input` without having to move any attributes to it (only set `type` and `slot`)

2. Refactor test suites properly to test against shadow / slotted input 

3. Refactor observers for propagating `required`, `invalid` etc (make them DRY)

4. Document that only subset of supported `input` properties is propagated to the slotted input

5. Emphasise using native input properties, like `placeholder` (it does not make sense to propagate them). Follow the same idea as we previously did for renderers, to replace all the demos eventually

6. Allow adding the input lazily. It's not guaranteed that the `<input>` is included inside the `<vaadin-text-field>` when the text field gets connected/initialized.

Prototype at #274

jmart1:
How exactly do we make use of this from our Java code?  I want to utilize this phone example:

```
        <vaadin-text-field label="Phone">
          <input type="tel" slot="input" pattern="[0-9]{3}-[0-9]{3}-[0-9]{4}" placeholder="Format: 111-222-3333">
        </vaadin-text-field>
```

but from within my Java code.  Is that possible?

jmart1:
Came up with a way to accomplish setting the type to tel.

```
                UI.getCurrent().getPage().executeJs(
                        "document.getElementById(\"" + id + "\").shadowRoot\n" +
                                ".querySelector(\".vaadin-text-field-container\")\n" +
                                ".getElementsByTagName(\"div\")[0]\n" +
                                ".getElementsByTagName(\"slot\")[1]\n" +
                                ".getElementsByTagName(\"input\")[0]\n" +
                                ".setAttribute(\"type\", \"tel\");"
                );
```

where the variable id is the id on the TextField component.

PR #283, opened 2018-11-29
Slotted input

yuriy-fix:
Connects to #281

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/283)


demo/text-field-custom-demos.html
```diff
@@ -0,0 +1,78 @@
...
+      and include styles into the same scope as `vaadin-text-field`:
+      <code>&lt;custom-style&gt;&lt;style include="lumo-placeholder-styles"&gt;&lt;/style&gt;&lt;/custom-style&gt;</code> OR 
+      <code>&lt;custom-style&gt;&lt;style include="material-placeholder-styles"&gt;&lt;/style&gt;&lt;/custom-style&gt;</code>
+    </p>
+    <vaadin-demo-snippet id="text-field-custom-demos-phone">
+      <template preserve-content>
+        <custom-style>
+          <style include="lumo-placeholder-styles"></style>
+        </custom-style>
+        <vaadin-text-field label="Name">
```
tomivirkki: The demo component should have a `placeholder` as the description note is about styling it

demo/text-field-custom-demos.html
```diff
@@ -0,0 +1,78 @@
...
+  <template>
+    <style include="vaadin-component-demo-shared-styles">
+      :host {
+        display: block;
+      }
+    </style>
+
+    <h3>Custom Input</h3>
+    <p>
+      <b>Note:</b> In order to apply lumo or material theme styles to the placeholder
```
tomivirkki: There should preferably be some intro for the feature. Starting with NOTE is kind of unexpected.

demo/text-field-custom-demos.html
```diff
@@ -0,0 +1,78 @@
...
+      of the slotted input you should explicitly import a file
+      `theme/lumo/vaadin-placeholder-styles.html` or `theme/material/vaadin-placeholder-styles.html`
```
tomivirkki: Better not use backticks in the demo descriptions as they are not parsed as Markdown.

demo/text-field-custom-demos.html
```diff
@@ -0,0 +1,78 @@
...
+    </vaadin-demo-snippet>
+
+    <h3>Number Input</h3>
+    <p>
+      <b>Note:</b> All properties and attributes that `vaadin-text-field` has should be defined on the host.
+    </p>
+    <vaadin-demo-snippet id="text-field-custom-demos-phone">
+      <template preserve-content>
+        <vaadin-text-field label="Amount" pattern="^[1-9]{1,2}[0]?|100" prevent-invalid-input value="0">
+          <input type="number" slot="input" min="0" max="100" step="2">
```
tomivirkki: Wondering why the input allows any text on Safari. The "Prevent Invalid Input" example under "Validation" properly prevents typing letters on Safari also.

theme/material/vaadin-placeholder-styles.html
```diff
@@ -0,0 +1,66 @@
+<link rel="import" href="../../../vaadin-material-styles/color.html">
+
+<dom-module id="material-placeholder-styles">
+  <template>
+    <style>
+      vaadin-text-field > input[slot="input"]::-webkit-input-placeholder,
```
tomivirkki: The selectors assume the host tag name, so it doesn't work with extensions.

demo/text-field-custom-demos.html
tomivirkki: With intro I meant something like: "A native `<input>` element can optionally be defined separately as a child element for the `<vaadin-text-field>`. This will enable use-cases where having the `<input>` element available in the global DOM is necessary, such as supporting password managers or native forms.

The current intro, while valid, doesn't really tell the reader what "Custom Input" mentioned in the demo title means.

tomivirkki: There seems to be a nice intro in the `vaadin-text-field.html` docs already. Why not use that here also?

demo/text-field-custom-demos.html
tomivirkki: I still see lots of backticks in the demos

tomivirkki: Ping @yuriy-fix

PR #287, opened 2018-12-05
Align Lumo imports for password-field and text-area

web-padawan:
Currently, when importing `theme/lumo/vaadin-text-area.html`, you also get the `vaadin-text-field` component definition, although technically you might not need it. 

This change is unlikely to affect someone, as both components are still there in the same repo, and this structure is already implemented for Material.

In general, the users should import the components they are using explicitly, without relying on them being imported somewhere else (even in case of password-field extending text-field).

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/287)


web-padawan:
We probably should check all the components and see if there are cases like this elsewhere.

web-padawan:
Closing this. Will reopen if we continue new development in the existing repos after V14. Otherwise if we decide to move the components to monorepo, will keep that in mind during migration.

PR #290, opened 2018-12-16
Add <vaadin-email-field> component

samiheikki:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/290)

### DIFF
--- a/src/vaadin-email-field.html
+++ b/src/vaadin-email-field.html
@@ -1,0 +1,56 @@
+<!--
+@license
+Copyright (c) 2018 Vaadin Ltd.
+This program is available under Apache License Version 2.0, available at https://vaadin.com/license/
+-->
+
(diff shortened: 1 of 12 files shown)


src/vaadin-email-field.html
```diff
@@ -0,0 +1,67 @@
...
+    class EmailFieldElement extends Vaadin.TextFieldElement {
+      static get is() {
+        return 'vaadin-email-field';
+      }
+
+      static get version() {
+        return '2.3.0-alpha3';
+      }
+
+      static get template() {
```
tomivirkki: Not sure why the template override is required?

samiheikki: Oh yeah I forgot that there. I'll remove it

PR #291, opened 2018-12-21
Add visualisation when invalid input is prevented

### LINKED ISSUE #84: Add visualization when invalid input is prevented
Currently, when `prevent-invalid-input` in enabled and user tries, for example, to type letters in a numbers-only field, nothing happens.

UX is poor, we should somehow indicate that input was prevented. Without the indication, the user thinks that his keyboard is broken or there's some virus in his lovely Android phone.

Indication by default could be red border (shadow-box) blink for a short moment right after typing.

(This is the DX tests finding)

emarc:
Fixes #84 

Implementation temporarily (200ms, debounced) toggles a
"input-prevented" attribute. The default style colors the value (and
caret) red (using error color in Lumo/Material).
Text color was chosen, because the field might be "invalid", in which
case the regular "invalid" style would not be seen. The drawback is that
only the caret is colored if the field is empty.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/291)


src/vaadin-text-field-mixin.html
```diff
@@ -307,6 +307,13 @@
         const input = this.focusElement;
         if (input.value.length > 0 && !this.checkValidity()) {
           input.value = this.value || '';
+          // add input-prevented attribute for 200ms
+          this.setAttribute('input-prevented', '');
+          this._inputDebouncer = Polymer.Debouncer.debounce(
```
web-padawan: There should be 2 imports added on top of the HTML in this file:

```html
<link rel="import" href="../../polymer/lib/utils/async.html">
<link rel="import" href="../../polymer/lib/utils/debounce.html">
```

web-padawan:
@jouni we would like to get your approval for Lumo change, so PTAL.

theme/lumo/vaadin-text-field-styles.html
```diff
@@ -217,6 +217,10 @@
         box-shadow: 0 0 0 2px var(--lumo-error-color-50pct);
       }
 
+      :host([input-prevented]) [part="input-field"] {
+        color: var(--lumo-error-color);
```
jouni: Should use `--lumo-error-text-color` to ensure proper contrast.

theme/material/vaadin-text-field-styles.html
```diff
@@ -186,6 +186,10 @@
         background-color: var(--material-error-color);
       }
 
+      :host([input-prevented]) [part="input-field"] {
+        color: var(--material-error-color);
```
jouni: Should use `--material-error-text-color` to ensure proper contrast.

emarc:
Made requested changes - not sure if I should ping here, resolve conversation, or something else...

jouni:
> not sure if I should ping here, resolve conversation, or something else

I think this was the expected process, that reviewers resolve the conversations. Personally, I’ve disabled almost all Github notifications, so it’s useful to ping me explicitly at least 😄

limonte:
Released in https://github.com/vaadin/vaadin-text-field/releases/tag/v2.3.0-alpha6

PR #292, opened 2018-12-22
Improve vaadin-number-field

samiheikki:
Connects to https://github.com/vaadin/vaadin-text-field/issues/296

This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/292)

### DIFF
--- a/src/vaadin-number-field.html
+++ b/src/vaadin-number-field.html
@@ -39,6 +39,7 @@ This program is available under Apache License Version 2.0, available at https:/
       }
 
       [part="value"] {
+        /* Older Firefox versions (v47.0) requires !important */
         -moz-appearance: textfield !important;
       }
     </style>
@@ -195,11 +196,15 @@ This program is available under Apache License Version 2.0, available at https:/
         }
 
         _valueChanged(newVal, oldVal) {
-          if (!isNaN(newVal) && !isNaN(parseFloat(newVal).toFixed(this.__decimals))) {
-            this.value = parseFloat(newVal).toFixed(this.__decimals).toString();
-          } else {
-            this.value = this.focusElement.value = '';
+          // Validate value to be numeric
+          if (newVal && isNaN(parseFloat(newVal).toFixed(this.__decimals))) {
+            this.value = '';
+          } else if (!isNaN(parseFloat(this.value)) &&
+            parseFloat(this.value) !== parseFloat(parseFloat(this.value).toFixed(this.__decimals))) {
+            // Validate correct decimals
+            this.value = parseFloat(parseFloat(this.value).toFixed(this.__decimals));
           }
+
           super._valueChanged(this.value, oldVal);
         }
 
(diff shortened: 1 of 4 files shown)


src/vaadin-number-field.html
```diff
@@ -201,6 +212,17 @@
           }
           return super.checkValidity();
         }
+
+        _onInput(e) {
```
tomivirkki: Can't type negative number because the input is canceled when I type "-"

tomivirkki: Chrome allows me to type "eee" while Safari doesn't.

jouni:
Yes, I agree with Tomi.

src/vaadin-number-field.html
samiheikki: What Chrome version are you using? I'm using v71 and I can't type (or paste) "e".

samiheikki: Fixed.

samiheikki:
Removed input focus when using the stepper controls.

samiheikki:
As I've been implementing this I have noticed that trying to validate the `<input type="number">` value is surprisingly difficult.

`input.value` is not the same as the actual letters/numbers I'm typing to the input field (as shown in the demo) https://jsfiddle.net/samiheikki/q7dx8Lfw/12/ Try running it in Chrome & Safari

And thus I don't know what actually is typed in the number field when I'm relying on the browser's validated value of the number field.

So instead of using `<input type="number">` & custom validation I'm seeing 2 alternative options.

`<input type="text">` with our custom validation.
- `input.value` works as we expect it to.
- Easy to validate numbers and special characters like `,`, `.`, `-`, `e`
- Hurts a11y

`<input type="number">` with no custom validation.
- We can use `input.value` and that will always give us the correct numeric value
- Some browsers allow letters (but they are not added to the `input.value`)
- Works more like native `<input type="number">`

I'd be in favor of option 2. BTW I haven't implemented either of these yet.

@tomivirkki LMKWYT. I can discuss more about this.

tomivirkki:
2. sounds like the better option

src/vaadin-number-field.html
tomivirkki: Works for me now also

src/vaadin-number-field.html
```diff
@@ -151,9 +160,9 @@
         __add(sign) {
           const incr = sign * (this.step || 1);
           // Behave like native number input adjusting to the next exact multiple of step.
-          this.value = (incr + (incr * Math.floor((parseFloat(this.value || 0) / incr).toFixed(1)))).toFixed(this.__decimals);
+          this.value = this.focusElement.value =
+            (incr + (incr * Math.floor((parseFloat(this.value || 0) / incr).toFixed(1)))).toFixed(this.__decimals);
           this.dispatchEvent(new CustomEvent('change', {bubbles: true}));
-          this.focusElement.focus();
```
tomivirkki: This didn't work however, the field still gets focused on first time the button is clicked so I get the virtual keyboard (on both iOS and Android) which goes away the second time the button is clicked

src/vaadin-number-field.html
```diff
@@ -31,21 +39,23 @@
       }
 
       [part="value"] {
-        -moz-appearance: textfield;
+        -moz-appearance: textfield !important;
```
tomivirkki: Why `!important`?

src/vaadin-number-field.html
```diff
@@ -201,6 +226,28 @@
           }
           return super.checkValidity();
         }
+
+        _onInput(e) {
+          const input = this.focusElement;
+          if (e.data === '-' && (this.__previousValidInput === '-' || !this.__previousValidInput)) {
+            // Allow minus sign (-) as the first character
+            this.__previousValidInput = '-';
```
tomivirkki: None of the logic inside this method is tested!

tomivirkki:
@samiheikki UX of the last number field on the demo page (with steps) is pretty unexpected when using keyboard (both desktop and mobile). Can something be done about it?

kapture 2019-01-14 at 15 02 13

src/vaadin-number-field.html
samiheikki: Removed this logic as I implemented the "option 2".

src/vaadin-number-field.html
samiheikki: Added comment. Without it visual tests fail with Firefox.

samiheikki:
@tomivirkki That could be achieved by adding `!this.__userInput` to this if clause: https://github.com/vaadin/vaadin-text-field/blob/finalize-feature/vaadin-number-field/src/vaadin-number-field.html#L202

But then it won't validate the step when input value changes. So not sure which is more correct behaviour.

Issue #288, opened 2018-12-07
Autoselect feature


emarc:
### Description
Quite often, if would be desirable to (auto)select the content of the textfield when it is focused, in order to improve UX. 
The most convenient way to achieve this would be to add a `autoselect` attribute, similar to `autofocus` etc.

### Expected outcome
`<vaadin-text-field autoselect>` would cause the input text to be selected automatically when the field is focused. If the users starts typing, the existing content is replaced. User can press arrow-right to append stuff. And so on – you now what I'm talking about :-)

### Actual outcome
No easy way to achieve this ATM. 

### Related
#214 is a request for a selection API.

limonte:
Released in https://github.com/vaadin/vaadin-text-field/releases/tag/v2.3.0-alpha4 🎉

PR #294, opened 2019-01-03
Add autoselect

emarc:
Add `autoselect` – automatically select content when focused #288

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/294)

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -398,12 +398,10 @@ This program is available under Apache License Version 2.0, available at https:/
       if (slotted && !this._slottedInput) {
         this._validateSlottedValue(slotted);
         this._addInputListeners(slotted);
-        this._addAutoselectListener(slotted);
         this._addIEListeners(slotted);
         this._slottedInput = slotted;
       } else if (!this._slottedInput) {
         this._removeInputListeners(this._slottedInput);
-        this._removeAutoselectListener(this._slottedInput);
         this._removeIEListeners(this._slottedInput);
         this._slottedInput = undefined;
       }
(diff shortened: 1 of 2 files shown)


src/vaadin-text-field-mixin.html
```diff
@@ -471,20 +481,31 @@
...
+      node.removeEventListener('focus', this._boundAutoselect);
+    }
+
     ready() {
       super.ready();
 
       this._boundOnInput = this._onInput.bind(this);
       this._boundOnChange = this._onChange.bind(this);
       this._boundOnBlur = this.validate.bind(this);
+      this._boundAutoselect = this._autoselect.bind(this);
```
web-padawan: I would suggest using `this._boundOnFocus` to align with other listeners here.

src/vaadin-text-field-mixin.html
```diff
@@ -517,6 +538,12 @@
       this.value = '';
     }
 
+    _autoselect() {
```
web-padawan: I would suggest using `_onFocus` here to make it clear that this function is supposed to be event listener only, and distinguish from a potential "private" method.

src/vaadin-text-field-mixin.html
```diff
@@ -471,20 +481,31 @@
       node.removeEventListener('blur', this._boundOnBlur);
     }
 
+    _addAutoselectListener(node) {
```
web-padawan: It would be better to reuse existing `_addInputListeners` and `_removeInputListeners` methods above, and add corresponding lines there.

test/text-field.html
```diff
@@ -216,6 +216,24 @@
             });
           });
         }
+
+        describe(`autoselect ${condition}`, function() {
+          it('default value of autoselect should be false', function() {
+            expect(textField.autoselect).to.be.false;
+          });
```
web-padawan: nit: please add newline between tests (after each test except last one) as in the blocks above to keep formatting consistent

src/vaadin-text-field-mixin.html
emarc: I considered `focus`, but was a bit worried `_onFocus` would clash with something. I'll make the change, I agree it's much better, should have asked before :-)

emarc: Done.

src/vaadin-text-field-mixin.html
emarc: Done.

src/vaadin-text-field-mixin.html
emarc: Done.

test/text-field.html
emarc: Done.

web-padawan:
We forgot to merge into branch first and the tests are failing in Safari 9 😰 
I will investigate the failures and try to fix them while we are still in alpha.

emarc:
> We forgot to merge into branch first and the tests are failing in Safari 9 😰

For future reference: what should I have done to avoid this? Apparently `polymer test` does not run al the required tests?

web-padawan:
@emarc that's our team who should handle it. For PRs from forks the `push` build running tests on SauceLabs does not start, as Travis does not expose credentials to forks for security reasons.

So, we previously agreed within a team that we will merge PRs from forks into feature branch first, in order to check if everything is good with the SauceLabs tests.

Same applies to updating screenshots for visual tests (though we can coordinate, in case if you need SauceLabs credentials to run visual tests locally).

PR #300, opened 2019-01-16
fix: ensure slotchange input listeners removal does not throw

web-padawan:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/300)

### DIFF
--- a/test/text-field.html
+++ b/test/text-field.html
@@ -55,6 +55,40 @@
 
     });
 
+    describe('slots', () => {
+      let textField;
+
+      function afterFlush() {
+        return new Promise(resolve => {
+          if (window.ShadyDOM) {
+            window.ShadyDOM.flush();
+            resolve();
+          } else {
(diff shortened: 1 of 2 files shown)


test/text-field.html
```diff
@@ -55,6 +55,37 @@
...
+      });
+
+      it('should not break when adding unexpected content to input slot', done => {
+        const div = document.createElement('div');
+        div.setAttribute('slot', 'input');
+        textField.appendChild(div);
+        const errorSpy = sinon.spy();
+        window.onerror = errorSpy;
+        afterFlush().then(() => {
+          expect(errorSpy).to.not.be.called;
```
tomivirkki: Should probably call `errorSpy.restore();` before the expect so `window.onerror` doesn't remain patched

web-padawan: Done. For some reason there is no `restore()` method, so re-assigned `onerror` manually.

PR #303, opened 2019-01-21
Fix setting min/max value with controls

samiheikki:
Fixes "With `<vaadin-number-field has-controls min="2"></vaadin-number-field>`, the + button will set the value to 1 first (same problem with max="-2")"

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/303)

### DIFF
--- a/src/vaadin-number-field.html
+++ b/src/vaadin-number-field.html
@@ -165,22 +165,65 @@ This program is available under Apache License Version 2.0, available at https:/
         }
 
         _decreaseValue() {
-          this._allowed(-1) && this.__add(-1);
+          const incrementSign = this._getAllowedIncrementSign(-1);
+
+          if (incrementSign) {
+            this.__add(incrementSign);
+          }
         }
 
         _increaseValue() {
-          this._allowed(1) && this.__add(1);
+          const incrementSign = this._getAllowedIncrementSign(1);
+
+          if (incrementSign) {
+            this.__add(incrementSign);
+          }
         }
 
         __add(sign) {
           const incr = sign * (this.step || 1);
           // Behave like native number input adjusting to the next exact multiple of step.
-          this.value = this.focusElement.value =
-            (incr + (incr * Math.floor((parseFloat(this.value || 0) / incr).toFixed(1)))).toFixed(this.__decimals);
+          this.value = this.focusElement.value = this._getValue(incr);
           this.dispatchEvent(new CustomEvent('change', {bubbles: true}));
         }
 
-        _allowed(sign, value, min, max) {
+        _getValue(incr) {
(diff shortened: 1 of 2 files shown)


src/vaadin-number-field.html
```diff
@@ -165,22 +165,65 @@
...
+            } else if (!validMax) {
+              return this.max / (this.step || 1);
+            }
+          }
+        }
+
+        _allowed(sign) {
           const incr = sign * (this.step || 1);
           return !this.disabled && (incr < 0
             ? this.min == null || this.value > this.min
```
web-padawan: The check `this.value  > this.min` needs to be fixed. Currently the following scenario is not covered:

1. set value to `null`
2. set `min` to negative number, e.g. -10
3. `null > -10` here returns true
4. the "minus" control is not disabled

src/vaadin-number-field.html
```diff
@@ -165,22 +165,65 @@
...
+
+            if (validMin && validMax) {
+              return sign;
+            } else if (!validMax) {
+              return false;
+            } else if (!validMin) {
+              return this.min / initialIncr;
+            }
+          } else {
+            const validMin = this.min == null || this.value > this.min;
```
web-padawan: The `this.value > this.min` check will fail here as well if `value` is null and `min` is negative.

test/number-field.html
```diff
@@ -199,6 +199,40 @@
...
+        it('should increase value to min value with step when increaseButton is clicked', () => {
+          numberField.step = 0.1;
+          numberField.min = 0.5;
+
+          increaseButton.click();
+
+          expect(numberField.value).to.be.equal('0.5');
+        });
+
+        it('should decrease value to max value when decreaseButton is clicked', () => {
```
web-padawan: Please also add a test confirming the following:

1. set `this.min` to a negative number, e.g. `-10`
2. make sure the "minus" control is disabled
3. make sure that once "plus" is pressed, number field has -10 as value (currently it will be 0)

src/vaadin-number-field.html
samiheikki: I don't understand why that is expected behaviour.

1. I set value to `null` and to me it feels like `null` should behave like the value is `0` which it does.
2. If `min` is `-10` then from empty value I should be able to use the decrease button to go from empty (0) value to -10.
3. And then I should be able to click decrease button until the value is -10 at which point the decrease button is disabled

screenshot 2019-01-23 at 16 51 54

web-padawan: Discussed in person with Tomi and I agree with your point.
Let's consider `null` as 0 and implicit default value as also 0, as it is now.

PR #304, opened 2019-01-21
Prevent setting value over max and under min limit

samiheikki:
Fixes "Typing 3, then 2 will result in invalid appearance for a number field with max of 31"

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/304)

### DIFF
--- a/src/vaadin-number-field.html
+++ b/src/vaadin-number-field.html
@@ -209,6 +209,12 @@ This program is available under Apache License Version 2.0, available at https:/
         }
 
         __onInputChange() {
+          if (this.min && this.value < this.min) {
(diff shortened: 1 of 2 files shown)


tomivirkki:
Ah, the idea wasn't to automatically fix the value but to prevent the field from going invalid while the user is typing. This occurred during UX tests but can't seem to be able to reproduce now. @web-padawan let's close this PR if this isn't reproducible

tomivirkki:
Apparently this was because `preventInvalidInput` was also enabled for the field for UX. In that case the behaviour is as expected. Closing.

PR #307, opened 2019-01-28
Fix value control behaviour when value is out of range

### LINKED ISSUE #306: vaadin-number-field controls behave poorly if value is out of range
### Description
Decrease and increase value buttons do not work if as expected if value is out of valid range.

### Live Demo
Can be reproduced in [cdn demo](https://cdn.vaadin.com/vaadin-text-field/2.3.0-alpha8/demo/index.html#number-field-demos)

### Steps to reproduce
In `Number field with value limits` demo type 20 as value, and start "decreasing" the value with the decrease button value will go to 30, 40 and so on, even though max value is 10.

samiheikki:
Fixes #306 

Visual tests are failing for an unrelated issue 🤷‍♂️

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/307)

### DIFF
--- a/src/vaadin-number-field.html
+++ b/src/vaadin-number-field.html
@@ -166,25 +166,40 @@ This program is available under Apache License Version 2.0, available at https:/
         }
 
         _decreaseValue() {
-          const incrementSign = this._getAllowedIncrementSign(-1);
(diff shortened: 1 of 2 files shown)


web-padawan:
Visual tests are now fixed on master, rebased to restart the tests.

PR #314, opened 2019-02-11
Update element version numbers

samiheikki:
Add correct version to `vaadin-email-field`and `vaadin-number-field`

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/314)


samiheikki:
Visual tests are failing but should not be affected by this

PR #320, opened 2019-02-15
Fix text-area scroll with fixed height in material theme

### LINKED ISSUE #310: vaadin-text-area and scroll in Material theme
### Description
When using Material Theme and  vaadin-tex-area has a defined height, writing text and when the scroll is show, the Material bar in the Bottom is scrolled together with the Text.

text_area_scroll

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Android Chrome

niiyeboah:
Fixes #310

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/320)

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -189,7 +189,8 @@ This program is available under Apache License Version 2.0, available at https:/
          */
         errorMessage: {
           type: String,
-          value: ''
+          value: '',
+          reflectToAttribute: true
         },
 
         /**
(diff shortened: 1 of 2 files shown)


samiheikki:
With this fix the error label is displayed above the material line when it should be under it.

height-max

tomivirkki:
This could probably be fixed by moving the underline from `[part="input-field"]` to `.vaadin-text-area-container`

niiyeboah:
> This could probably be fixed by moving the underline from `[part="input-field"]` to `.vaadin-text-area-container`

That will cause the same issue shown in the screenshot @samiheikki posted. The red underline is underneath the error message.

PR #329, opened 2019-02-25
Remove preventInput flag on requestAnimationFrame

### LINKED ISSUE #328: Prevent invalid input broken on IE11
After https://github.com/vaadin/vaadin-text-field/pull/326 you can actually type an invalid character into the text field (once after focus) even if the field is marked `preventInvalidInput`.

yuriy-fix:
Fixes #328

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/329)

### DIFF
--- a/test/text-field.html
+++ b/test/text-field.html
@@ -353,7 +353,7 @@
 
           textField.dispatchEvent(new Event('focusin'));
 
-          setTimeout(() => {
+          requestAnimationFrame(() => {
             input.value = 'a';
             const inputEvent = new Event('input');
             input.dispatchEvent(inputEvent);
--- a/test/text-field.html
+++ b/test/text-field.html
@@ -347,6 +347,22 @@
           });
         });
 
+        (IE ? it : it.skip)('should prevent invalid if input event is user originated', done => {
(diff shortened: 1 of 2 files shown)


tomivirkki:
No test?

yuriy-fix:
Have come up with an idea of the test with the scenario described in the ticket.

test/text-field.html
```diff
@@ -347,6 +347,22 @@
           });
         });
 
+        (IE ? it : it.skip)('should prevent invalid if input event is user originated', done => {
+          textField.preventInvalidInput = true;
+          textField.pattern = '[0-9]*';
+
+          textField.dispatchEvent(new Event('focusin'));
+
+          setTimeout(() => {
```
tomivirkki: The fix uses raf but the test uses timeout. That sounds like a recipe for a flakey test.

yuriy-fix: Good catch! Replaced with `requestAnimationFrame`.

yuriy-fix:
Removing flag in the `onInput` as well.

PR #330, opened 2019-02-25
fix: align text-field properly in material theme

web-padawan:
Connected to vaadin/vaadin-custom-field#23

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/330)


web-padawan:
Note: the relevant code is from vaadin/vaadin-themes@099eeeed301c921cbb4692e6c9bfdc324a299f98 which is almost 2 years ago, and the original issue has been most likely fixed in Chrome (and I can't reproduce it anymore).

PR #333, opened 2019-02-27
Fix number-field value and step issues

### LINKED ISSUE #331: Setting both value and step in number field truncates a valid value
### Description
When setting both `value` and `step` with decimals the initial value is modified even though it's valid

### Expected outcome
Value is not modified and `9.99` is shown

### Actual outcome
Value is modified and shown as `10.0`

### Demo
The following snippet can be used to reproduce the issue
`<vaadin-number-field value="9.99" step="0.01" label="value and step"></vaadin-number-field>`
image

niiyeboah:
- Allow programmatically setting any value regardless of step
- Change clicking + or - to increment or decrement value to the next multiple of step (offset by the min)
- Change checkValidity to check when value is not multiple of step (offset by the min)
- Change value to always return a string

Fixes #331

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/333)

### DIFF
--- a/src/vaadin-number-field.html
+++ b/src/vaadin-number-field.html
@@ -208,7 +208,18 @@ This program is available under Apache License Version 2.0, available at https:/
         }
 
         _getValue(incr) {
-          return (incr + (incr * Math.floor((parseFloat(this.value || 0) / incr).toFixed(1)))).toFixed(this.__decimals);
+          let result = this.min || incr;
+          if (this.value) {
+            const number = parseFloat(this.value);
+            let nextMultiple = parseFloat((Math.ceil(number / incr) * incr).toFixed(this.__decimals));
+            if (this.min) {
+              let minOffset = this.min % this.step;
+              minOffset = incr < 0 ? -minOffset : minOffset;
+              nextMultiple = this.min ? nextMultiple + minOffset : nextMultiple;
+            }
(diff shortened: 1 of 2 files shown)


tomivirkki:
There's still an inconsistency between the two first examples. If you type 1.5 to the fist number field, its `validate()` returns `true`, but if you type the same value to the second field (required), `validate()` returns `false` (and the component appears invalid).

src/vaadin-number-field.html
```diff
@@ -289,8 +302,9 @@
 
         checkValidity() {
           // text-field mixin does not check against `min` and `max`
-          if (this.min !== undefined || this.max !== undefined) {
-            this.invalid = !this.inputElement.checkValidity();
+          if (this.value && (this.min !== undefined || this.max !== undefined || this.step)) {
+            const stepBase = this.min || 0;
+            this.invalid = this.value < this.min || this.value > this.max || (this.value - stepBase).toFixed(this.__decimals) % this.step !== 0;
```
tomivirkki: According to https://caniuse.com/#feat=input-number iOS 9 Safari wouldn't take min/max/step into account in validation, but just tested it manually and it _seems_ to work. See if we could after all just rely on the native input's validation.

niiyeboah: Ok I see. I changed back to using the native validation. Will the Travis tests catch if this doesn't work?

PR #341, opened 2019-03-10
Prevent invalid number-field value when step is over min/max

### LINKED ISSUE #340: Number field: Increment buttons can produce invalid value
If I set the `<vaadin-number-field>`'s
- value to 1
- max to 2
- step to 10
it still shows the plus button as enabled, and clicking it will set the value to 10

samiheikki:
Fixes #340

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/341)

### DIFF
--- a/test/number-field.html
+++ b/test/number-field.html
@@ -251,6 +251,26 @@
 
           expect(numberField.value).to.be.equal(numberField.min);
         });
+
+        it('should decrease value to min value when step is over the min value', () => {
+          numberField.value = 2;
+          numberField.min = 1;
+          numberField.step = 10;
+
+          decreaseButton.click();
+
+          expect(numberField.value).to.be.equal('1');
+        });
+
+        it('should increase value to max value when step is over the max value', () => {
+          numberField.value = 1;
+          numberField.max = 2;
+          numberField.step = 10;
+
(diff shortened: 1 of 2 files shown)


tomivirkki:
After this change, the increase button behaves very unexpectedly. Could also use a rebase on top of the new master since https://github.com/vaadin/vaadin-text-field/pull/333 was merged

numberfield-increment

test/number-field.html
```diff
@@ -251,6 +251,26 @@
...
+        });
+
+        it('should increase value to max value when step is over the max value', () => {
+          numberField.value = 1;
+          numberField.max = 2;
+          numberField.step = 10;
+
+          increaseButton.click();
+
+          expect(numberField.value).to.be.equal('2');
```
tomivirkki: I don't think this is the right assumption. If step is 10, then 2 would be an invalid value. See how native `<input type="number">` works in this situation and align the behaviour.

samiheikki:
Putting this on hold and waiting for https://github.com/vaadin/vaadin-text-field/pull/343 to be merged first.

tomivirkki:
Closed by #343

PR #363, opened 2019-04-20
Increase touch target for number field buttons

### LINKED ISSUE #359: Number field buttons have too small touch target size
The size of the buttons is only 20x20 pixels. Should be at least twice that to be comfortable to use on a touch device, which is the primary reason for using the buttons anyways.

Currently, it’s easy to misclick, and as a side-effect focus the input field which pops up the software keyboard which in turn moves the field on the screen (as seen in #125 ).

samiheikki:
Fixes #359

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/363)


samiheikki:
@web-padawan Fixed.

PR #366, opened 2019-05-02
fix: clear button related bug

### LINKED ISSUE #365: Programmatically setting the value after clicking the clear button doesn't work properly
### Description
If you click the clear button (or call `textField.$.clearButton.click();`) and then programmatically set the value `textField.value = 'foo';` the component will be in an inconsistent state as the value of the native `<input>` element will be empty `textField.inputElement.value === ''` but the value of text field will be as expected `textField.value === 'foo'`.

This isn't a big issue when using `vaadin-text-field` directly but it's a bigger issue for many components which internally use `vaadin-text-field` with `clear-button-visible` API (e.g. `vaadin-combo-box`).

### Expected outcome
`textField.value` and `textField.inputElement.value` should always be in sync.
In this case both should be equal to `'foo'`.

`textField.value === textField.inputElement.value`

### Actual outcome
`textField.value !== textField.inputElement.value`
`textField.value` === `'foo'`
`textField.inputElement.value` === `''`

### Live Demo
"Display the Clear Button" demo in
https://cdn.vaadin.com/vaadin-text-field/2.4.1/demo/#text-field-basic-demos

### Steps to reproduce
1. Select/inspect the `vaadin-text-field` element (of the clear button demo) in dev tools.
2. `$0.value = 'foo';`
3. `$0.$.clearButton.click();`
4. Notice that `$0.value` returns `'foo'` but `$0.inputElement.value` returns `''` and the text field looks empty but it still shows the clear button.

### Browsers Affected
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE 11
- [ ] iOS Safari
- [ ] Android Chrome

Tested in Chrome, probably affects other browsers also.

Haprog:
Programmatically setting the value after clicking the clear button didn't work properly making the input look like empty but displaying the clear button and the component still had a value. This caused more easily noticeable issues with other components that internally use vaadin-text-field (when using the clear button).

Fixes #365

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/366)

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -364,7 +364,22 @@ This program is available under Apache License Version 2.0, available at https:/
           return;
         }
       }
-      this.__userInput = true;
+
+      if (this.__inputFromClearButton) {
+        this.__inputFromClearButton = false;
+        // In this case when input event is triggered by clear button we need to skip
+        // setting "this.__userInput = true" because the clear button clears the value
+        // first triggering _valueChanged() before this input event. If we wouldn't
+        // skip this, then __userInput would be left true until the next _valueChanged()
+        // happens which would then prevent setting this.inputElement.value correctly
+        // leaving it out of sync with this.value.
+      } else {
+        // This is used to prevent setting this.inputElement.value in _valueChanged()
+        // when this input event was actually triggered by the user from the input
+        // element itself (thus the value there should be correct already).
+        // This flag is set back to false in _valueChanged().
+        this.__userInput = true;
+      }
       this.value = e.target.value;
     }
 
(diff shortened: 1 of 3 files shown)


Haprog:
CI tests failed once for IE11 only (random failure) and when running for second time IE11 passed but one of the iOS suites timed out (random failure) so in total everything passed at least once. I still restarted the failing job once more, but I consider this ready for review.

Haprog:
This fix has been released in [v2.4.2](https://github.com/vaadin/vaadin-text-field/releases/tag/v2.4.2) (issue was a regression in v2.4.0 so didn't affect earlier versions).

PR #367, opened 2019-05-07
Add a "label-suffix" slot

### LINKED ISSUE #212: Add support for a label "suffix" slot

heruan:
This is my second attempt to fix #212 (the first was #213); it is cleaner and can easily be applied to other components porting the style to a mixin in `lumo-styles`.

As for your previous @tomivirkki considerations:

- this is a small breaking change, so I do not expect it to be merged before 3.0
- this would not affect the Material theme, so it would be a feature of Lumo

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/367)

### DIFF
--- a/demo/text-field-label-demos.html
+++ b/demo/text-field-label-demos.html
@@ -1,0 +1,39 @@
+<dom-module id="text-field-label-demos">
+  <template>
+    <style include="vaadin-component-demo-shared-styles">
+      :host {
+        display: block;
+      }
+    </style>
+
+
+    <h3>Label suffix</h3>
+    <vaadin-demo-snippet id="text-field-label-demos-label-suffix">
+      <template preserve-content>
+        <vaadin-password-field id="password" label="New password">
+          <vaadin-button id="suggest" slot="label-suffix" theme="tertiary-inline">suggest</vaadin-button>
+        </vaadin-password-field>
+        <script>
+          window.addDemoReadyListener('#text-field-label-demos-label-suffix', function(document) {
+            const field = document.getElementById('password');
+            const button = document.getElementById('suggest');
+            button.addEventListener('click', function(event) {
+              field.value = Math.random().toString(36).substring(7);
+              field._setPasswordVisible(true);
+            });
+          });
+        </script>
+      </template>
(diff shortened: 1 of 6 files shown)


demo/text-field-label-demos.html
```diff
@@ -0,0 +1,39 @@
...
+        display: block;
+      }
+    </style>
+
+
+    <h3>Label suffix</h3>
+    <vaadin-demo-snippet id="text-field-label-demos-label-suffix">
+      <template preserve-content>
+        <vaadin-password-field id="password" label="New password">
+          <vaadin-button id="suggest" slot="label-suffix" theme="tertiary-inline">suggest</vaadin-button>
```
web-padawan: Per MDN web docs, placing interactive content inside a `<label>` is a bad practice we should avoid.
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Interactive_content

> Don't place interactive elements such as anchors or buttons inside a label. Doing so makes it difficult for people to activate the form input associated with the label.

heruan: Thank you @web-padawan for the reference, I already paid attention to this and if you look closely there is no interactive content inside the `<label>` element: the slot is side-by-side the label, inside a container `<div>`.

web-padawan: Thanks, missed that.

web-padawan:
Closing per https://github.com/vaadin/vaadin-text-field/issues/212#issuecomment-494898284

PR #369, opened 2019-05-09
fix: make it possible to detect clear button as the source for input …

### LINKED ISSUE #368: Make it possible to detect the source of the 'input' and 'change' events triggered by the clear button
### Description
Currently (since https://github.com/vaadin/vaadin-text-field/pull/353 in v2.4.0+) triggering the clear button sends an `input` event. This has caused a regression in vaadin-combo-box#805. To fix that, we need to be able to tell when an `input` event was triggered by the clear button (instead of user typing).

The closest we can get now is to check if the `value` of `vaadin-text-field` is empty, but that can also mean that the user just removed all the text using the keyboard (backspace or del) instead of the clear button, and the clear button needs special logic/behaviour at least in the case of `vaadin-combo-box` (and similar components) where we don't want the click of a clear button to open the dropdown (but regular input should).

Currently it seems that only the `input` event needs this for fixing the above mentioned regression, but we might as well add the same feature to the `change` event. And imo these should be added as new public API that is documented (basically a small new feature) though this is not strictly necessary to fix the regression (if we want to make it as a private API instead).

This could be implemented e.g. by setting a custom property on these events (after they've been created), but there might be some better way.

Blocks vaadin/vaadin-combo-box#805.

### Expected outcome
When an event listener for `input` or `change` event on `vaadin-text-field` is run, it should be possible to somehow check if that event was triggered via the clear button or not.

### Actual outcome
It doesn't seem to be possible to detect if the event was triggered via the clear button click handler since the clear button creates new events and dispatches them on the native input element.

Haprog:
…and change events

Add property `__fromClearButton = true` (private API for now) to the `input` and `change` events sent by the clear button.

Fixes #368 and unblocks vaadin/vaadin-combo-box#805.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/369)

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -365,19 +365,7 @@ This program is available under Apache License Version 2.0, available at https:/
         }
       }
 
-      if (this.__inputFromClearButton) {
-        this.__inputFromClearButton = false;
-        // In this case when input event is triggered by clear button we need to skip
-        // setting "this.__userInput = true" because the clear button clears the value
-        // first triggering _valueChanged() before this input event. If we wouldn't
-        // skip this, then __userInput would be left true until the next _valueChanged()
-        // happens which would then prevent setting this.inputElement.value correctly
-        // leaving it out of sync with this.value.
-      } else {
-        // This is used to prevent setting this.inputElement.value in _valueChanged()
-        // when this input event was actually triggered by the user from the input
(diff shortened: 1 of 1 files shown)


Haprog:
Travis build failed only because SauceLabs removed support for testing on iOS 9.3 (all other tests pass).

PR #373, opened 2019-05-15
Fix textField demos

shadikhani:
Fix textField demos

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/373)


demo/text-field-validators-demos.html
```diff
@@ -49,20 +58,46 @@ <h3>Synchronous Custom Validator</h3>
...
       then override the <code>checkValidity()</code> method to validate the user’s input.</p>
-
-    <vaadin-demo-snippet id="text-field-validators-demos-synchronous-custom-validator" ignore-ie when-defined="vaadin-text-field">
+    <vaadin-demo-snippet id="text-field-validators-demos-synchronous-custom-validator"
+      ignore-ie when-defined="vaadin-text-field">
       <template preserve-content>
-        <employee-role-text-field label="Position"></employee-role-text-field>
-        Valid options: Architect, Developer or Designer
+        <employee-role-text-field label="ID" minlength="10" maxlength="10"></employee-role-text-field>
+        <p>Valid ID: A 10 digit number that sum of the digits are dividable to 10. For
```
tomivirkki: Should be rephrased

demo/text-field-validators-demos.html
```diff
@@ -28,19 +29,27 @@ <h3>Minlength and Maxlength</h3>
 
 
     <h3>Pattern</h3>
-    <p>The <code>pattern</code> property specifies a regular expression that the element’s value is
+    <p>The <code>pattern</code> property specifies a regular expression that the element’s
+      value is
```
tomivirkki: Accidental line break?

demo/text-field-validators-demos.html
```diff
@@ -117,14 +156,13 @@ <h3>Asynchronous Custom Validator</h3>
...
         </script>
-
       </template>
     </vaadin-demo-snippet>
 
-  <dom-module id="my-ajax">
-    <template>
-    </template>
-    <script>
+    <dom-module id="my-ajax">
```
Haprog: This element doesn't seem to be used anywhere anymore so you should remove this whole `<dom-module id="my-ajax">`.

demo/text-field-validators-demos.html
```diff
@@ -81,13 +115,11 @@ <h3>Asynchronous Custom Validator</h3>
...
-
-    <vaadin-demo-snippet id="text-field-validators-demos-asynchronous-custom-validator" ignore-ie when-defined="vaadin-text-field">
+    <vaadin-demo-snippet id="text-field-validators-demos-asynchronous-custom-validator"
+      ignore-ie when-defined="vaadin-text-field">
       <template preserve-content>
-        <my-ajax id="ajax1" handle-as="json" url="https://vaadin.com/validate-element"></my-ajax>
-        <chemical-element-text-field label="Element Name"></chemical-element-text-field>
-        Type a valid Chemical Element such as Hydrogen, Oxygen, Carbon, etc.
-
+        <chemical-element-text-field label="Username"></chemical-element-text-field>
```
Haprog: This element should be renamed to something like `username-field` and also the class name from `ChemicalElement` to `UsernameField` because it doesn't deal with chemical elements anymore.

demo/text-field-validators-demos.html
shadikhani: What about this?
Valid ID: Use a 10 digit number. The sum of digits must be dividable by 10. 
For example 1111111111.

demo/text-field-validators-demos.html
shadikhani: I used "format document" of the VS code that has changed the format. I will use the previous format for the texts.

demo/text-field-validators-demos.html
Haprog: I thought "dividable" isn't even a word, but after googling it I learned it seems to be ok to use that word in general. Though I would still recommend to use the much more common form "divisible" instead since it doesn't look so strange. And in this context when talking about divisibility of numbers (math) it's actually more correct to use "divisible" anyway since that implies it's "evenly divisible" which is what is meant here.

I think the updated one now is ok, but I would write it like this instead (and maybe add another example of a valid number):
`Valid ID: A 10 digit integer where the sum of all digits is divisible by 10. For example 1111111111 or 1234000000.`

demo/text-field-validators-demos.html
```diff
@@ -49,20 +54,46 @@ <h3>Synchronous Custom Validator</h3>
...
     <p>Extend <code>Vaadin.TextFieldElement</code> to create your own custom element,
       then override the <code>checkValidity()</code> method to validate the user’s input.</p>
-
-    <vaadin-demo-snippet id="text-field-validators-demos-synchronous-custom-validator" ignore-ie when-defined="vaadin-text-field">
+    <vaadin-demo-snippet id="text-field-validators-demos-synchronous-custom-validator"
+      ignore-ie when-defined="vaadin-text-field">
       <template preserve-content>
-        <employee-role-text-field label="Position"></employee-role-text-field>
-        Valid options: Architect, Developer or Designer
+        <employee-role-text-field label="ID" minlength="10" maxlength="10"></employee-role-text-field>
```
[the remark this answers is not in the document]
Haprog: This is not relevant anymore as I pushed an update (for simplifying things) where we just use `vaadin-text-field` here without unnecessarily extending it.

Haprog:
I pushed an update where I simplified the two custom validator demos so that they don't unnecessarily create a new custom element extending `vaadin-text-field` just for setting the validator function. This makes the code shorter (reduces boilerplate) and also makes these examples work on IE11.

Haprog:
CI seems to fail only on:
```
firefox 67 (error)                    
Error: Error: [get("http://localhost:8081/components/[secure]/generated-index.html?cli_browser_id=1")] Error response status: 13, UnknownError - An unknown server-side error occurred while processing the command. Selenium error: Failed to decode response from marionette
Build info: version: '3.12.0', revision: '7c6e0b3', time: '2018-05-08T15:15:08.936Z'
System info: host: 'travis-job-dace60cc-baf3-4c59-aa3b-5fb7e291ab9d', ip: '127.0.1.1', os.name: 'Linux', os.arch: 'amd64', os.version: '4.4.0-101-generic', java.version: '1.8.0_151'
```

But all tests pass locally on Firefox 67.

PR #391, opened 2019-06-26
fix(material): ensure long label does not overflow

### LINKED ISSUE #388: [material] label overflow
### Description
A label that is longer than the text field itself breaks out of the container.

### Expected outcome
The label should be truncated. This happens already in the lumo theme and in the similar [paper-input element](https://www.webcomponents.org/element/@polymer/paper-input).

### Actual outcome
The label is as long as it's text content and overlays neighbouring elements.

### Live Demo
Screenshot from 2019-06-19 16-43-16

### Steps to reproduce
1. Put a `vaadin-text-field` element in the page with a label longer than it's width.

### Browsers Affected

- [x] Chrome
_tested only in chrome, but any browser should be affected_

### Possible Fix
Inspired by the [paper-input element](https://www.webcomponents.org/element/@polymer/paper-input).
Add this snippet to `theme/material/vaadin-text-field-styles.html`:
```
:host([has-label]:not([has-value]):not([focused]):not([invalid]):not([theme~="always-float-label"])) [part="label"] {
    ...
    width: 100%;
}

[part="label"] {
    ...
    width: 133%;
}
```

web-padawan:
Fixes #388

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/391)


web-padawan:
Thanks, will add the test for `always-float-label` also, and look into `transition`.

PR #393, opened 2019-06-29
Add rows and cols propertie

### LINKED ISSUE #351: Allow setting "rows" and "cols" for TextArea
A `textarea`-element has the properties `rows` and `cols`, which allow you to specify the visible number of lines and the visible width of the `textarea`.
The `vaadin-textarea` doesn't offer an API to set those values for the nested `textarea`-element.
The issue #170 might be related to this, but it is about setting min/max height, which is already possible using CSS.

samiheikki:
Fixes #351

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/393)


src/vaadin-text-area.html
```diff
@@ -183,8 +204,10 @@
           this._oldValueLength = valueLength;
 
           const inputHeight = input.scrollHeight;
-          if (inputHeight > input.clientHeight) {
+          if (inputHeight > input.clientHeight && !this.rows) {
             input.style.height = inputHeight + 'px';
+          } else if (this.rows) {
+            input.style.height = 'initial';
```
tomivirkki: Looks like a `<vaadin-text-area>` with `rows` doesn't scroll when the content is long enough to overflow. This is due to the explicit `overflow: hidden` on the `<textarea>` which should probably not be in use when `rows` is defined.

yuriy-fix:
The targeted issue is closed now: https://github.com/vaadin/vaadin-text-field/issues/351#issuecomment-512137496

PR #394, opened 2019-06-29
Move readonly state border from pseudo element to input-field part

### LINKED ISSUE #228: Lumo animations are strange in the vaadin-grid CRUD demo example
When changing the `readonly` property from `true` to `false` the strange blinking is happening:

peek 2018-06-01 13-38

Is that bug or feature?

Chrome/Linux, Firefox/Linux

jouni:
Fixes #228

This is one way to fix the issue. It’s not really pretty (more `calc()`s, not really DRY), but it works.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-text-field/394)

### DIFF
--- a/theme/lumo/vaadin-text-field-styles.html
+++ b/theme/lumo/vaadin-text-field-styles.html
@@ -38,8 +38,8 @@
       /* Slotted by vaadin-select-text-field */
       [part="input-field"] ::slotted([part="value"]) {
         cursor: inherit;
-        min-height: var(--lumo-text-field-size);
-        padding: 0 0.25em;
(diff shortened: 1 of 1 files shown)


web-padawan:
It has been a while since this PR was submitted. Apologies about keeping it open for so long without any progress. 

The component has been converted to Polymer 3 so there is a merge conflict. 
Also, the new development for Vaadin web components will happen in the [monorepo](https://github.com/vaadin/vaadin-web-components). 

I'm closing this now. We will submit a new PR to the monorepo when the linked issue is prioritised.

PR #399, opened 2019-08-02
chore(wct): Update wct config

Haprog:
Make Saucelabs tests pass again


Haprog:
Needed to get `2.3` branch green again after cherry picking fixes from `master` to it.

PR #401, opened 2019-08-07
fix: stuck in invalid state after last validation constraint is removed while invalid

### LINKED ISSUE #400: Stuck in invalid state after last validation constraint is removed while invalid
### Description

The validation considers `required`, `pattern`, `maxlength` and `minlength`. If any of those are set, then it sets `invalid` boolean based on the return value of `checkValidity()` on the native `<input>`. If none of those are set you can still manually set `invalid` to `true` or `false` based on your own logic and `vaadin-text-field` keeps the invalid state as it has been set.

Currently the problem seems to be that when you only have one of those validation constraints set (e.g. `required`) and the field has been validated and become invalid (e.g. by just focusing and blurring a field that was required), and then you remove the constraint, it doesn't run `checkValidity()` on the native input anymore and it keeps the `invalid` state (as if it had been added manually). Basically the text field never automatically removes `invalid` state when there are no active validation constraints.

This causes bad behaviour when e.g. `required` property has been bound to some value that updates based on other controls (e.g. a checkbox which toggles if some fields are required or not).

I think the proper logic here would be to observe when any of `required`, `pattern`, `maxlength` or `minlength` constraints is removed _and_ `invalid == true` already, then set `this.invalid = this.inputElement.checkValidity();`

When the field is `invalid` and any validation constraint (`required`, `pattern`, `maxlength`, `minlength`) is removed, validation should be run again and `invalid` state updated accordingly (even if there are no more active validation constraints). If it was the last/only validation constraint (e.g. changing a required field to not required) `invalid` should basically be set to `false` so that it doesn't get stuck as invalid.

This was first reported at vaadin/vaadin-combo-box#830 and might affect also other components that internally use `vaadin-text-field`.

### Steps to reproduce
Put this on a page:

```html
<vaadin-text-field required="{{tfRequired}}"></vaadin-text-field><br>
<label>Required <input type="checkbox" checked="{{tfRequired::change}}"></label>
```

1. toggle checkbox on (makes text field required)
2. focus and unfocus the text field (makes it invalid since it's still empty)
3. toggle checkbox off (makes text field not required)
4. write something in the text field and unfocus it, notice that text field is still shown as invalid

Now the only way to make it valid again is to either manually set `invalid = true` or make it required again and then change the value to something other than empty.

Haprog:
Fixes #400

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -317,9 +317,17 @@ This program is available under Apache License Version 2.0, available at https:/
         '_hostAccessiblePropsChanged(' + HOST_PROPS.accessible.join(', ') + ')',
         '_getActiveErrorId(invalid, errorMessage, _errorId)',
         '_getActiveLabelId(label, _labelId, _inputId)',
-        '__observeOffsetHeight(errorMessage, invalid, label)',
-        '__constraintsChanged(required, minlength, maxlength, pattern, min, max, step)'
+        '__observeOffsetHeight(errorMessage, invalid, label)'
       ];
+    }
+
+    constructor() {
+      super();
+      // This complex observer needs to be added dynamically here (instead of defining it above in the `get observers()`)
+      // so that it runs after complex observers of inheriting classes. Otherwise e.g. `_stepOrMinChanged()` observer of
+      // vaadin-number-field would run after this and the `min` and `step` properties would not yet be propagated to
+      // the `inputElement` when this runs.
+      this._createMethodObserver('__constraintsChanged(required, minlength, maxlength, pattern, min, max, step)');
     }
 
     get focusElement() {
@@ -508,7 +516,9 @@ This program is available under Apache License Version 2.0, available at https:/
         return;
       }
 
-      if (!required && !minlength && !maxlength && !pattern && !min && !max) {
+      const isNumUnset = n => (!n && n !== 0);
+
+      if (!required && !minlength && !maxlength && !pattern && isNumUnset(min) && isNumUnset(max)) {
         this.invalid = false;
       } else {
         this.validate();
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -317,7 +317,8 @@ This program is available under Apache License Version 2.0, available at https:/
         '_hostAccessiblePropsChanged(' + HOST_PROPS.accessible.join(', ') + ')',
         '_getActiveErrorId(invalid, errorMessage, _errorId)',
         '_getActiveLabelId(label, _labelId, _inputId)',
-        '__observeOffsetHeight(errorMessage, invalid, label)'
+        '__observeOffsetHeight(errorMessage, invalid, label)',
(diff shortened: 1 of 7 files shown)


tomivirkki:
Quite a lot of changes in this PR. All the tests seem to pass with the following additional observer to `vaadin-text-field-mixin` (could be extended by numberfield if -mixin shouldn't know about min, max, step):

```
__constraintsChanged(required, minLength, maxLength, pattern, min, max, step) {
  if (!required && !minLength && !maxLength && !pattern && !min && !max) {
    this.invalid = false;
  }
  if (this.invalid) {
    this.validate();
  }
}
```
Am I missing something essential? If so, there should be more tests to cover all the changes.

Haprog:
> Quite a lot of changes in this PR. All the tests seem to pass with the following additional observer to `vaadin-text-field-mixin` (could be extended by numberfield if -mixin shouldn't know about min, max, step):
> 
> ```
> __constraintsChanged(required, minLength, maxLength, pattern, min, max, step) {
>   if (!required && !minLength && !maxLength && !pattern && !min && !max) {
>     this.invalid = false;
>   }
>   if (this.invalid) {
>     this.validate();
>   }
> }
> ```
> 
> Am I missing something essential? If so, there should be more tests to cover all the changes.

The problem with this simpler approach is that it will run validation only when all those are undefined, but not when there is still some active constraint but one or more other constraints were unset.

It's a shame that the multi dependency observers in Polymer don't provide the old values so you don't know which of those dependencies changed without manually tracking the old values elsewhere. The single property observer provides old value but I also ran to some issue when thinking about using individual observers.

I'll add a test case where there are many constraints applied and one is unset.

Haprog:
After taking a closer look, the simple `__constraintsChanged` observer approach seemed to work good after all considering all the cases I could think of.

I rebased this branch to latest master (with the flaky test fixes) and updated to use this simplified approach. Also added one more test and fixed some random typos in the test files.

src/vaadin-text-field-mixin.html
```diff
@@ -502,6 +503,18 @@
       }
     }
 
+    __constraintsChanged(required, minlength, maxlength, pattern, min, max, step) {
+      if (!this.invalid) {
+        return;
+      }
+
+      if (!required && !minlength && !maxlength && !pattern && !min && !max) {
```
tomivirkki: Oh, `min` and `max` can be set to 0 which is falsy value, but still a valid constraint. We should probably check for `undefined`/`null` here instead.

Haprog: Good catch. I'm sure I was thinking about this but forgot it while changing other things.

Atm it makes an invalid field valid if e.g. `max` is set to `0` and value is bigger than `0`. I'll add test cases for these min/max 0 validity changes and fix this.

Haprog: Fixing this revealed that the test `should update "invalid" state when "step" is removed` was a false positive before (it started failing). Fixed that test too and pushed an update.

Haprog: Or well the test was correct, but was passing due to broken behaviour. Fixing that behaviour just revealed that another fix was needed too.

PR #410, opened 2019-09-11
Add helper text API and slot

niiyeboah:
- Added `helper-text` part below text field
- Added observer to add `has-helper` attribute when `helperText` property is set
- Added slotchange event handler to add `has-helper` attribute when content placed in `helper` slot
- Added helper text to `aria-describedby`
- Added observer for error message to add `has-error-message` attribute (used to hide helper in material theme when error message is set)


src/vaadin-text-field-mixin.html
```diff
@@ -445,6 +457,33 @@
       }
     }
 
+    _hintChanged(hint) {
+      if (hint !== '' && hint !== null) {
```
tomivirkki: Looks like text-field-mixin has `_setOrToggleAttribute` that you could probably use in these three functions instead of having to repeat the logic

src/vaadin-text-field-mixin.html
```diff
@@ -571,6 +610,12 @@
       this.shadowRoot.querySelector('[name="input"], [name="textarea"]')
         .addEventListener('slotchange', this._onSlotChange.bind(this));
 
+      this._slottedHint = this.querySelector(`[slot="hint"]`);
```
tomivirkki: Isn't this same logic as in `_hintSlotChange`? You could invoke `this._hintSlotChange()` here instead.

src/vaadin-text-field-mixin.html
```diff
@@ -450,32 +450,20 @@
     }
 
     _labelChanged(label) {
-      if (label !== '' && label != null) {
-        this.setAttribute('has-label', '');
-      } else {
-        this.removeAttribute('has-label');
-      }
+      this._setOrToggleAttribute('has-label', label === 0 ? true : label, this);
```
tomivirkki: label is of type string so `label === 0` can never be true unless a developer illegally assigns a value of wrong type

niiyeboah: I had to check for that scenario because of [this](https://github.com/vaadin/vaadin-text-field/blob/f5cc5da639abaed7f5145a00958bf7a910118b8a/test/text-field.html#L301) test.

tomivirkki: :) Oh lord. We actually have tests to verify that the property works with invalid value types  🤦‍♂ ...while the property is even documented to be of type String. I guess it's safer now to keep the current behaviour. This will get fixed when we switch to TypeScript.

src/vaadin-text-field-mixin.html
niiyeboah: Done

src/vaadin-text-field-mixin.html
tomivirkki: It could be shortened to `label === 0 || !!label` however

tomivirkki:
There's an alignment issue if you place the helper above the field.

Screenshot 2019-09-30 11 01 27

src/vaadin-text-field-mixin.html
```diff
@@ -680,10 +714,16 @@
...
+    _getActiveErrorId(invalid, errorMessage, errorId, helperText, helperTextId) {
+      let ids = '';
+      if (helperText) {
+        ids += helperTextId;
+      }
+      if (errorMessage && invalid) {
+        ids += helperText ? ' ' : '';
+        ids += errorId;
+      }
+      this._setOrToggleAttribute('aria-describedby', ids, this.focusElement);
```
tomivirkki: A slotted helper text doesn't seem to label the focusElement. Is this intentional (as slotted content might not be text)?

src/vaadin-text-field-mixin.html
```diff
@@ -571,6 +601,9 @@
       this.shadowRoot.querySelector('[name="input"], [name="textarea"]')
         .addEventListener('slotchange', this._onSlotChange.bind(this));
 
+      this._onHelperSlotChange();
```
tomivirkki: Can this be tested?

src/vaadin-text-field-mixin.html
```diff
@@ -438,10 +450,28 @@
...
+      this._slottedHelper = this.querySelector(`[slot="helper"]`);
+      const slottedNodes = this.root.querySelector(`[name="helper"]`).assignedNodes();
+
+      // Only not empty if not a text node
+      // Text nodes are added by the helperText prop and not the helper slot
+      // The filter is added due to shady DOM triggering this slotchange event on helperText prop change
+      if (slottedNodes.filter(node => node.nodeType !== 3).length) {
+        this.setAttribute('has-helper', 'slotted');
+      } else if (this.helperText === '' || this.helperText === null) {
+        this.removeAttribute('has-helper');
```
tomivirkki: Untested code

src/vaadin-text-field-mixin.html
```diff
@@ -189,7 +189,8 @@
          */
         errorMessage: {
           type: String,
-          value: ''
+          value: '',
+          observer: '_errorMessageChanged'
```
tomivirkki: Untested code

src/vaadin-text-field-mixin.html
niiyeboah: Ah I totally missed this. I changed it to add the label on slotchange.

I think if slotted content has no text then maybe it is up to the dev to handle the a11y issues?

Tested here: [1](https://github.com/vaadin/vaadin-text-field/blob/4a20bf4aa9f84b881367d33740bf14a4bf7bc36e/test/accessibility.html#L327).

src/vaadin-text-field-mixin.html
niiyeboah: Tested here: [1](https://github.com/vaadin/vaadin-text-field/blob/4a20bf4aa9f84b881367d33740bf14a4bf7bc36e/test/text-field.html#L358).

The test fails with shady DOM if this line is removed.

src/vaadin-text-field-mixin.html
niiyeboah: Tested here: [1](https://github.com/vaadin/vaadin-text-field/blob/4a20bf4aa9f84b881367d33740bf14a4bf7bc36e/test/text-field.html#L358), [2](https://github.com/vaadin/vaadin-text-field/blob/4a20bf4aa9f84b881367d33740bf14a4bf7bc36e/test/accessibility.html#L327).

src/vaadin-text-field-mixin.html
niiyeboah: Tested here: [1](https://github.com/vaadin/vaadin-text-field/blob/4a20bf4aa9f84b881367d33740bf14a4bf7bc36e/test/text-field.html#L387)

niiyeboah: Done

src/vaadin-text-field-mixin.html
niiyeboah: Done

src/vaadin-text-field-mixin.html
niiyeboah: Done

src/vaadin-text-field-mixin.html
niiyeboah: Done

theme/material/vaadin-text-field-styles.html
```diff
@@ -146,12 +146,30 @@
         opacity: 0;
       }
 
+      /* According to matrial theme guidelines, helper text should be hidden when error message is set and input is invalid */
```
yuriy-fix: Nit: `material`

test/text-field.html
```diff
@@ -326,6 +334,70 @@
...
+          });
+
+          it('setting number helper updates has-helper attribute', function() {
+            textField.helperText = 0;
+            expect(textField.hasAttribute('has-helper')).to.be.true;
+          });
+
+          it('text-field with slotted helper updates has-helper attribute', function(done) {
+            const textFieldSlottedHelper = fixture('default-with-slotted-helper');
+            setTimeout(() => {
```
yuriy-fix: I guess we can avoid timeouts by using similar approach as in `text-field` with dispatching `slot-change`:

https://github.com/vaadin/vaadin-text-field/blob/2c9c8aac97e16646d247d6646be7979770903541/test/custom-input.html#L79

niiyeboah:
Thanks Yuriy :)

I fixed the error message position and also the label padding when the helper text is above (I agree it looks better):

Screenshot 2019-10-04 at 13 17 38

When I test the voice over, the helper text is read after the label, and then if there is an error message it will be read after that. There is however some weird behaviour which happens on my system. When the helper is a slot it seems to cause the page to reload sometimes if you have voice over enabled and focus on the element.

Screenshot 2019-10-04 at 13 12 16

Screenshot 2019-10-04 at 13 12 28

test/text-field.html
niiyeboah: Done, except it didn't work with this test: https://github.com/vaadin/vaadin-text-field/blob/7cae1f2e3e6a3d7646bcabc80486a0b713a0f8fa/test/text-field.html#L379-L388

theme/material/vaadin-text-field-styles.html
niiyeboah: Done

src/vaadin-text-field-mixin.html
```diff
@@ -438,10 +452,29 @@
...
+    _helperTextChanged(helperText) {
+      this._setOrToggleAttribute('has-helper', helperText === 0 || !!helperText, this);
+    }
+
+    _errorMessageChanged(errorMessage) {
+      this._setOrToggleAttribute('has-error-message', errorMessage === 0 || !!errorMessage, this);
+    }
+
+    _onHelperSlotChange() {
+      this._slottedHelper = this.querySelector(`[slot="helper"]`);
```
tomivirkki: Unused code

src/vaadin-text-field-mixin.html
```diff
@@ -680,10 +717,17 @@
...
-        (errorMessage && invalid ? errorId : undefined),
-        this.focusElement);
+    _getActiveErrorId(invalid, errorMessage, errorId, helperText, helperTextId, hasSlottedHelper) {
+      let ids = '';
+      const hasHelper = helperText || hasSlottedHelper;
+      if (hasHelper) {
+        ids += helperTextId;
+      }
+      if (errorMessage && invalid) {
+        ids += hasHelper ? ' ' : '';
```
tomivirkki: nit: building a space separated string like this is a bit clumsy. Rather push to an array of strings and then use `ids.join(' ')` to form the space separated string.

src/vaadin-text-field-mixin.html
niiyeboah: Done

src/vaadin-text-field-mixin.html
niiyeboah: Done

test/text-field.html
tomivirkki: Hmm...for me the same approach works in that test also

PR #418, opened 2019-09-27
fix: fields should not remove initially set invalid status on ready

### LINKED ISSUE #417: Fields should not remove initially set invalid status on ready
### Description
`vaadin-email-field` and `vaadin-number-field` currently automatically set `invalid = false` when the component becomes ready (e.g. if it was originally set as invalid via `invalid` attribute).

This is a regression introduced in https://github.com/vaadin/vaadin-text-field/commit/f49fa9b62fa44150357ca5813930e490a6de8c6c (v2.4.10 and v2.3.9)

This doesn't seem to affect e.g. `vaadin-text-field` directly in simple cases but does affect it in cases where it is used as an internal component with data binding (e.g. in vaadin-date-picker).

### Expected outcome
All fields in the vaadin-text-field repository should keep the invalid state that has explicitly been set (if the developer doesn't explicitly modify other validation contraint properties/attributes after the component is ready).

### Actual outcome
`vaadin-email-field` and `vaadin-number-field` automatically become valid once the component is ready even if there was an `invalid` attribute set initially.

### Steps to reproduce
1. Put this on a page:
   ```html
   <vaadin-email-field invalid></vaadin-email-field>
   <vaadin-number-field invalid></vaadin-number-field>
   ```
2. When you load the page you see that invalid state flashes and returns to valid state.

Haprog:
Fixes #417.


Haprog:
@web-padawan Great solution, thanks! I was initially trying to do something similar but I couldn't figure out a way to override the observer defined in a superclass. I didn't think of creating it in another (overridable) method and calling it in `ready()`.

I fast-forwarded this PR branch to contain your changes as is.

web-padawan:
This needs to be cherry-picker to `2.4` branch and released (as master is already 2.5 alpha)

Haprog:
Cherry-picked to 2.4 and released as [v2.4.12](https://github.com/vaadin/vaadin-text-field/releases/tag/v2.4.12)

PR #421, opened 2019-10-07
fix: avoid using initial step in validation

### LINKED ISSUE #420: NumberField: setting min enables validation for step
Steps to reproduce:
1. Create a simple `<vaadin-number-field>` without any attributes/properties
2. Set the value to `1.5` (notice that the field is valid)
3. Set `min=0` for the field

Expected outcome: the field is still valid

Actual outcome: the field becomes invalid

When the step is not explicitly set by user, it's used just for the plus and minus buttons, but not for validation. Setting the step property enables the step-based validation, but it's unexpected that setting a min-constraint does the same.

These lines reveal the issue:
https://github.com/vaadin/vaadin-text-field/blob/7599194862bd40fba47ab7b84254934a60de6087/src/vaadin-number-field.html#L345-L347

I suggest that instead of asking the native input element to validate the field, add the trivial implementation of min/max-validation here. Only if the step has changed, validate via the native input.

Another problem is that setting `step=1` doesn't trigger the step-based validation, but `step="1"` does. This logic should be based on whether user has set the property, not based on its absolute value.

pekam:
setting the min or max property should not trigger step-based validation

fix #420

### DIFF
--- a/src/vaadin-number-field.html
+++ b/src/vaadin-number-field.html
@@ -120,7 +120,6 @@ This program is available under Apache License Version 2.0, available at https:/
              */
             step: {
               type: Number,
-              reflectToAttribute: true,
               value: 1,
               observer: '_stepChanged'
             }
@@ -303,13 +302,13 @@ This program is available under Apache License Version 2.0, available at https:/
         }
 
         _stepChanged(step) {
-          if (!this.__stepChangedCalled && step === 1) {
-            // Setting the initial value, avoid using it in validation.
-            this.inputElement.step = 'any';
-          } else {
-            this.inputElement.step = step;
-          }
+          // Avoid using initial value in validation
+          const validateByStep = this.__stepChangedCalled
+            || (!this.__stepChangedCalled && this.getAttribute('step') !== null);
+          this.inputElement.step = validateByStep ? step : 'any';
+
           this.__stepChangedCalled = true;
+          this.setAttribute('step', step);
         }
 
         _minChanged(min) {
@@ -348,7 +347,7 @@ This program is available under Apache License Version 2.0, available at https:/
 
         checkValidity() {
           // text-field mixin does not check against `min`, `max` and `step`
-          if (this.min !== undefined || this.max !== undefined || this.step !== 1) {
+          if (this.min !== undefined || this.max !== undefined || this.inputElement.step !== 'any') {
             this.invalid = !this.inputElement.checkValidity();
           }
           return super.checkValidity();
--- a/src/vaadin-number-field.html
+++ b/src/vaadin-number-field.html
@@ -303,9 +303,9 @@ This program is available under Apache License Version 2.0, available at https:/
 
         _stepChanged(step) {
           // Avoid using initial value in validation
-          const validateByStep = this.__stepChangedCalled
+          this.__validateByStep = this.__stepChangedCalled
             || (!this.__stepChangedCalled && this.getAttribute('step') !== null);
-          this.inputElement.step = validateByStep ? step : 'any';
+          this.inputElement.step = this.__validateByStep ? step : 'any';
 
           this.__stepChangedCalled = true;
           this.setAttribute('step', step);
@@ -347,7 +347,7 @@ This program is available under Apache License Version 2.0, available at https:/
 
         checkValidity() {
           // text-field mixin does not check against `min`, `max` and `step`
-          if (this.min !== undefined || this.max !== undefined || this.inputElement.step !== 'any') {
+          if (this.min !== undefined || this.max !== undefined || this.__validateByStep) {
             this.invalid = !this.inputElement.checkValidity();
           }
           return super.checkValidity();
--- a/src/vaadin-number-field.html
+++ b/src/vaadin-number-field.html
@@ -303,7 +303,13 @@ This program is available under Apache License Version 2.0, available at https:/
         }
 
         _stepChanged(step) {
-          this.inputElement.step = step;
+          if (!this.__stepChangedCalled) {
+            // Setting the initial value, avoid using it in validation.
+            this.inputElement.step = 'any';
+            this.__stepChangedCalled = true;
+          } else {
+            this.inputElement.step = step;
+          }
         }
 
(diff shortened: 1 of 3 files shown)


pekam:
> If you set a custom step value with an attribute, the internal <input>'s step becomes "any"

Fixed by also checking `step === 1`. The only problem is that it doesn't recognize if user sets `step="1"` as attribute. In this case the step won't be used in validation (same as with the default step).

I think this is the best we can do. WDYT?

Haprog:
@pekam I think you could use `getAttribute('step')` to check if step has been set as an attribute to use it in validation (when attribute is set).

pekam:
@Haprog I've tested that. Unfortunately the `step` attribute has been updated on the first observer call, no matter if it was set as an attribute or property. So it's not possible to recognize if it was set originally as an attribute or property.

pekam:
If we consider the expected behavior to be:
_Step should be used in validation when user has set it as an attribute or property_,
then this PR in its current form fixes most cases, although not covering everything.

But the problem is that it will break one case which was previously working as expected:
- Setting `step="1"` as an attribute or `step=1` (as number) as property
- Also setting `min` and/or `max`

Expected: validated with step

Although this previously worked just because setting `min` or `max` triggers `step` validation, it's a case that would be a regression from this PR.

So I'm actually not sure if this can be merged. It is fixing more than it's breaking, but it is breaking something that previously worked.

Haprog:
I guess we could maybe remove `reflectToAttribute` and set the attribute manually at the end of the property change listener. Or can you think of some problems with that?

pekam:
That doesn't help in this case since it's initially defined as an attribute, which is reflected to property. It propagates this way even when `reflectToAttribute=false`.

I'm trying if I can use `attributeChangedCallback`.

Haprog:
> That doesn't help in this case since it's initially defined as an attribute, which is reflected to property. It propagates this way even when `reflectToAttribute=false`.

I don't think it matters in this case if it has been propagated from attribute to property already. If `reflectToAttribute` is false and in first step property change handler we check if the attribute is there we know that the developer explicitly set the attribute and validation should be.

Haprog:
Well I guess that would only solve the problem with setting it via attribute to `"1"` but it would still be an issue if the developer wanted to set it via the property to `1`. Maybe that could be documented as a known issue (until we can refactor to get rid of the explicit default value for `step`) so developer could workaround it by setting it as attribute instead of property if they need it initially set to `1` and to have validation.

pekam:
`reflectToAttribute` doesn't make any difference. Both the property and attribute are already updated when running the observer.

pekam:
But it seems that we can check if the attribute was set in `connectedCallback`, before Polymer does its magic.

Haprog:
`reflectToAttribute` should make the difference that it doesn't automatically reflect to attribute at all, right? As I see it that is the only thing needed to make it work as setting via attribute.

`reflectToAttribute: false`:
- first `_stepChanged` called where `(!this.__stepChangedCalled && step === 1)` equals true:
  - check `getAttribute('step')`, if attribute is found, use validation, user has explicitly set the attribute.
  - if no attribute is found at this point, just set input step to `any` like before
  - at the end `setAttribute('step', step)`

pekam:
D'oh, you're right. Sorry about my brain fart. :D Of course it makes a difference in the case when setting only the property but not the attribute.

Haprog:
So basically when `reflectToAttribute` is true, then in the step change handler the attribute is always set and we don't know if it was set by user or by polymer (from property), but when `reflectToAttribute` is false, in the first call to the step change we know that if the attribute is there, it was set by the user, and if it's not there it was not set by the user.

Haprog:
The harder case is if the user tries to set step to 1 (either via attribute or property) after the component is ready/initialized already. (since then the default value 1 is already in both property and attribute and change events might not happen if it's being set to the same value it's at already (I don't remember if it's the same case with setting attribute?).

Not sure if it would help to replace `step` "Polymer property" with setter and getter methods. I guess property setter would be called even if you set the same value to it again?

pekam:
I have tried overriding the `set step()` function but it doesn't get called. Polymer apparently overrides that. If the property is not declared in `properties`, then the setter is called. Trying to investigate it more.

That case at least has an easy workaround: setting the property as a string "1".

src/vaadin-number-field.html
```diff
@@ -303,13 +302,13 @@
...
         _stepChanged(step) {
-          if (!this.__stepChangedCalled && step === 1) {
-            // Setting the initial value, avoid using it in validation.
-            this.inputElement.step = 'any';
-          } else {
-            this.inputElement.step = step;
-          }
+          // Avoid using initial value in validation
+          const validateByStep = this.__stepChangedCalled
+            || (!this.__stepChangedCalled && this.getAttribute('step') !== null);
```
Haprog: This can be simplified to:
```js
const validateByStep = this.__stepChangedCalled || this.getAttribute('step') !== null;
```

Haprog: If it executes the right side of `||` we already know that `__stepChangedCalled == false`.

pekam: Thanks! Looks like my brain is on Monday-mode today :P

pekam: Done.

pekam:
> make a new major alpha version with that breaking change so this situation would improve and simplify the code when we can bump the new major version to stable and include in some later version of platform.

IMO it's not such a big deal that it would be worth to fix for the Polymer-based platform majors, which won't be long-term supported. But when rebuilding with Lit, this should be taken care of.

Haprog:
I guess so, but I'm mainly thinking if it's not a big effort to change it might be easier to do now so we won't forget about it and implement the same logic with default value when porting to Lit. It would be less refactoring required when doing the Lit version if this part was done already.

pekam:
Issue created #435. We can continue discussion on this elsewhere.

One more thing, @Haprog do you agree with my latest commit which I pushed after your approval? Read the full commit message for relevant info.

pekam:
Ok, thanks for the review! :)

PR #424, opened 2019-10-15
feat/refactor: add _enabledCharPattern property

pekam:
Protected feature extracted from `<vaadin-integer-field>` to allow
reusing the same logic for `BigDecimalField`.


pekam:
New testes were not added for this generic feature. The coverage of the existing `<vaadin-integer-field>` tests is enough.

PR #442, opened 2019-11-13
Check the number field validation constraints inside checkValidity

yuriy-fix:


src/vaadin-number-field.html
```diff
@@ -344,15 +344,16 @@
           this.validate();
         }
 
-        validate() {
+        checkValidity() {
```
yuriy-fix: Can be overridden as, not sure which one is easier to read 🤔:
```
checkValidity() {
  // text-field mixin does not check against `min`, `max` and `step`
  if (this.min !== undefined || this.max !== undefined || this.__validateByStep) {
    return this.inputElement.checkValidity();
  }
  
  return super.checkValidity;
}

tomivirkki: The shorter version you suggested looks nicer, not a blocker

PR #452, opened 2020-02-10
Add __forceCheckValidity and test

yuriy-fix:
Connected to https://github.com/vaadin/vaadin-date-picker/issues/696.

In many cases the component containing `text-field` (i.e.) faces a problem with the validation of internal input, because `checkValidity` of the `text-field` returns `false` based on the previous value of the `invalid` property. (`return !this.invalid;`) There wasn’t any simple way of making the field `checkValidity` without setting validation constraints. 

This PR adds this possibility.


test/validation.html
```diff
@@ -200,6 +200,15 @@
             expect(tf.invalid).to.be.true;
           });
 
+          it('should force check and change invalid property if `__forceCheckValidity` is set', function() {
```
web-padawan: Note, we have to have a correct test for this behavior without using private API in future.

Not a blocker but adding a TODO would not harm.

Haprog: I agree since this is only private now, there should be some note about it being needed for our other components (e.g. date-time-picker) so it doesn't accidentally get removed later. Should probably be converted to protected or public API later (preferably with some better naming or some docs).

yuriy-fix:
There is also possibility to go with:
```
checkValidity() {
  if (this.required || this.pattern || this.maxlength || this.minlength || this.__checkValidityForced) {
    this.__checkValidityForced = false;
    return this.inputElement.checkValidity();
  } else {
    return !this.invalid;
  }
}

__forceCheckValidity() {
  this.__checkValidityForced = true;
}
```

test/validation.html
yuriy-fix: Note added.

PR #461, opened 2020-03-25
fix: ARIA for disabled, required, readonly and invalid states

### LINKED ISSUE #430: vaadin-text-field's input has invalid aria-required attribute value
### Description
In required `vaadin-text-field`s and derived fields, `aria-required` attribute should be `true` for required fields.

### Expected outcome
https://www.w3.org/TR/WCAG20-TECHS/ARIA2.html
> The aria-required property can have values of "true" or "false". For example, if a user must fill in an address field, then aria-required is set to "true".
> Example 1: ` <input type="text" name="usrname" id="usrname" aria-required="true"/>[*]`

### Actual outcome
`aria-required` is a valueless (boolean) attribute

### Demo
https://vaadin.com/components/vaadin-text-field/html-examples/text-field-validators-demos

Haprog:
Do not set `aria-disabled`, `aria-required` or `aria-readonly` attributes since they are unnecessary when we're using the native HTML `disabled`, `required` and `readonly` attributes. Previously these three `aria-*` attributes were set with incorrect value (valid values for these are string `"true"` or `"false"`).

Use correct values for setting `aria-invalid` attribute (using string `"true"` when invalid).

Fixes #430.

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -117,8 +117,8 @@ This program is available under Apache License Version 2.0, available at https:/
 
   const HOST_PROPS = {
     default: ['list', 'autofocus', 'pattern', 'autocapitalize', 'autocorrect', 'maxlength',
-      'minlength', 'name', 'placeholder', 'autocomplete', 'title'],
-    accessible: ['disabled', 'readonly', 'required', 'invalid']
+      'minlength', 'name', 'placeholder', 'autocomplete', 'title', 'disabled', 'readonly', 'required'],
+    accessible: ['invalid']
   };
 
   const PROP_TYPE = {
@@ -499,10 +499,10 @@ This program is available under Apache License Version 2.0, available at https:/
       const input = this.inputElement;
       const attributeNames = HOST_PROPS[type];
 
-      if (type === 'accessible') {
+      if (type === PROP_TYPE.ACCESSIBLE) {
         attributeNames.forEach((attr, index) => {
           this._setOrToggleAttribute(attr, attributesValues[index], input);
-          this._setOrToggleAttribute(`aria-${attr}`, attributesValues[index], input);
+          this._setOrToggleAttribute(`aria-${attr}`, attributesValues[index] ? 'true' : false, input);
         });
       } else {
         attributeNames.forEach((attr, index) => {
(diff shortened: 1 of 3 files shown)


src/vaadin-text-field-mixin.html
```diff
@@ -117,8 +117,8 @@
 
   const HOST_PROPS = {
     default: ['list', 'autofocus', 'pattern', 'autocapitalize', 'autocorrect', 'maxlength',
-      'minlength', 'name', 'placeholder', 'autocomplete', 'title'],
-    accessible: ['disabled', 'readonly', 'required', 'invalid']
+      'minlength', 'name', 'placeholder', 'autocomplete', 'title', 'disabled', 'readonly', 'required'],
+    accessible: ['invalid']
```
Haprog: The `accessible` set is only a list of props that are also propagated to a corresponding `aria-*` attribute.

'disabled', 'readonly', 'required' were moved to the `default` set so they only propagate to the native HTML attribute of the same name on the input (without aria attribute).

src/vaadin-text-field-mixin.html
```diff
@@ -499,10 +499,10 @@
       const input = this.inputElement;
       const attributeNames = HOST_PROPS[type];
 
-      if (type === 'accessible') {
+      if (type === PROP_TYPE.ACCESSIBLE) {
         attributeNames.forEach((attr, index) => {
           this._setOrToggleAttribute(attr, attributesValues[index], input);
-          this._setOrToggleAttribute(`aria-${attr}`, attributesValues[index], input);
+          this._setOrToggleAttribute(`aria-${attr}`, attributesValues[index] ? 'true' : false, input);
```
Haprog: This makes it so that when setting `aria-*` attribute here (now only `invalid` -> `aria-invalid`) it is always cast to boolean and it either sets `aria-*="true"` or unsets the attribute when evaluating to false.

test/accessibility.html
```diff
@@ -221,21 +221,30 @@
...
 
             it('should not be marked required', function() {
-              expect(input.hasAttribute('aria-required')).to.be.false;
+              expect(input.hasAttribute('required')).to.be.false;
             });
 
             it('should be marked required', function() {
               tf.required = true;
-              expect(input.hasAttribute('aria-required')).to.be.true;
+              expect(input.getAttribute('required')).to.equal('');
```
Haprog: Explicitly checking for empty string value instead of using `hasAttribute` so that it would fail if the attribute would have some unexpected value instead.

PR #468, opened 2020-04-27
fix: make email field work like in LTR on RTL mode

### LINKED ISSUE #466: I don't expect the email field to be right aligned in RTL mode
### Description
It makes no sense to right align the text in email fields in RTL mode.

### Expected outcome
Text direction to be left to right.

### Actual outcome
Text direction is right to left.

### Live Demo
Tested with bookstore RTL sample.

### Steps to reproduce
1. Put a `vaadin-email-field` element in the page in RTL mode.
2. Type something into it.

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] IE 11
- [x] iOS Safari
- [x] Android Chrome

DiegoCardoso:
Fix #466

### DIFF
--- a/src/vaadin-email-field.html
+++ b/src/vaadin-email-field.html
@@ -14,6 +14,22 @@ This program is available under Apache License Version 2.0, available at https:/
       :host([dir="rtl"]) [part="input-field"] {
         direction: ltr;
       }
+
+      :host([dir="rtl"]) [part="value"]::placeholder {
+        direction: rtl;
+        text-align: left;
+      }
+
+      :host([dir="rtl"]) [part="input-field"] ::slotted(input)::placeholder {
+        direction: rtl;
+        text-align: left;
+      }
+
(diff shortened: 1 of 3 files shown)


src/vaadin-email-field.html
```diff
@@ -8,55 +8,97 @@
...
+      :host([dir="rtl"]) [part="value"]::placeholder {
+        direction: rtl;
+        text-align: right;
       }
 
-      ready() {
-        super.ready();
-        this.inputElement.type = 'email';
-        this.inputElement.autocapitalize = 'off';
+      :host([dir="rtl"]) [part="input-field"] ::slotted(input)::placeholder {
```
DiegoCardoso, line 23: Had to split this style with the one above, because this is not valid in Safari and it was not applying for the default input content. There's one issue [opened in the Webkit Bugzilla](https://bugs.webkit.org/show_bug.cgi?id=178237).

PR #470, opened 2020-04-29
Fix RTL styles for numeric fields

yuriy-fix:
Connected to https://github.com/vaadin/vaadin-text-field-flow/issues/302


yuriy-fix:
@shadikhani provided some clarifications about the issue, those are:
1. The value should be entered from LTR.
2. Placeholder should be RTL but its text-alignment should be left.
3. The buttons `+` and `-` should be the same as LTR

DiegoCardoso:
For LTR, the placeholder is center aligned when the field has the controls, but for RTL they are left aligned:
Screenshot 2020-04-30 at 12 02 35

Should it be still center aligned on this case?

shadikhani:
> For LTR, the placeholder is center aligned when the field has the controls, but for RTL they are left aligned:
> Screenshot 2020-04-30 at 12 02 35
> 
> Should it be still center aligned on this case?

For RTL it also should be center-aligned.

yuriy-fix:
@shadikhani , @DiegoCardoso , is it correct now?

Screen Shot 2020-04-30 at 13 08 40

shadikhani:
> @shadikhani , @DiegoCardoso , is it correct now?
> 
> Screen Shot 2020-04-30 at 13 08 40

Yes. That is perfect :)

DiegoCardoso:
It needs screenshot update now :)

src/vaadin-number-field.html
```diff
@@ -42,6 +42,28 @@
...
+        direction: rtl;
+      }
+
+      :host([dir="rtl"]) [part="value"]:-ms-input-placeholder,
+      :host([dir="rtl"]) [part="input-field"] ::slotted(input):-ms-input-placeholder {
+        direction: rtl;
+      }
+
+      :host([dir="rtl"]:not([has-controls])) [part="value"]::placeholder,
+      :host([dir="rtl"]:not([has-controls])) [part="input-field"] ::slotted(input)::placeholder {
```
DiegoCardoso, line 64: It is missing selector for IE11 placeholder.

PR #474, opened 2020-05-18
feat: prepare for TS defs generation

### LINKED ISSUE #467: Generate TypeScript definitions for P3
Child of https://github.com/vaadin/vaadin-core/issues/186

web-padawan:
Fixed #467 

The generated TS definitions look like this:

### `vaadin-email-field.d.ts`

```ts
import {TextFieldElement} from './vaadin-text-field.js';

import {DomModule} from '@polymer/polymer/lib/elements/dom-module.js';

declare class EmailFieldElement extends TextFieldElement {
  ready(): void;
  _createConstraintsObserver(): void;
}

declare global {
  interface HTMLElementTagNameMap {
    "vaadin-email-field": EmailFieldElement;
  }
}

export {EmailFieldElement};
```

### `vaadin-integer-field.d.ts`

```ts
import {NumberFieldElement} from './vaadin-number-field.js';

declare class IntegerFieldElement extends NumberFieldElement {
  ready(): void;
  _valueChanged(newVal: unknown|null, oldVal: unknown|null): void;
  _stepChanged(newVal: number, oldVal: number): void;
}

declare global {
  interface HTMLElementTagNameMap {
    "vaadin-integer-field": IntegerFieldElement;
  }
}

export {IntegerFieldElement};
```

### `vaadin-number-field.d.ts`

```ts
import {TextFieldElement} from './vaadin-text-field.js';

import {DomModule} from '@polymer/polymer/lib/elements/dom-module.js';

declare class NumberFieldElement extends TextFieldElement {
  hasControls: boolean;
  min: number|null|undefined;
  max: number|null|undefined;
  step: number;
  ready(): void;
  _createConstraintsObserver(): void;
  _valueChanged(newVal: unknown|null, oldVal: unknown|null): void;
  checkValidity(): boolean;
  _onKeyDown(e: KeyboardEvent): void;
  _stepChanged(newVal: number, oldVal: number): void;
}

declare global {
  interface HTMLElementTagNameMap {
    "vaadin-number-field": NumberFieldElement;
  }
}

export {NumberFieldElement};
```

### `vaadin-password-field.d.ts`

```ts
import {TextFieldElement} from './vaadin-text-field.js';

import {DomModule} from '@polymer/polymer/lib/elements/dom-module.js';

declare class PasswordFieldElement extends TextFieldElement {
  revealButtonHidden: boolean;
  readonly passwordVisible: boolean;
  ready(): void;
  _onChange(e: Event): void;
}

declare global {
  interface HTMLElementTagNameMap {
    "vaadin-password-field": PasswordFieldElement;
  }
}

export {PasswordFieldElement};
```

## `vaadin-text-area.d.ts`

```ts
import {PolymerElement} from '@polymer/polymer/polymer-element.js';

import {TextFieldMixin} from './vaadin-text-field-mixin.js';

import {ControlStateMixin} from '@vaadin/vaadin-control-state-mixin/vaadin-control-state-mixin.js';

import {ThemableMixin} from '@vaadin/vaadin-themable-mixin/vaadin-themable-mixin.js';

import {ElementMixin} from '@vaadin/vaadin-element-mixin/vaadin-element-mixin.js';

import {html} from '@polymer/polymer/lib/utils/html-tag.js';

declare class TextAreaElement extends
  TextFieldMixin(
  ControlStateMixin(
  ElementMixin(
  ThemableMixin(
  PolymerElement)))) {
  readonly _slottedTagName: string;
  ready(): void;
}

declare global {
  interface HTMLElementTagNameMap {
    "vaadin-text-area": TextAreaElement;
  }
}

export {TextAreaElement};
```

### `vaadin-text-field-mixin.d.ts`

```ts
export {TextFieldMixin};

declare function TextFieldMixin<T extends new (...args: any[]) => {}>(base: T): T & TextFieldMixinConstructor;

interface TextFieldMixinConstructor {
  new(...args: any[]): TextFieldMixin;
}

export {TextFieldMixinConstructor};

interface TextFieldMixin {
  readonly focusElement: HTMLElement|null;
  readonly _slottedTagName: string;
  autocomplete: string|null|undefined;
  autocorrect: string|null|undefined;
  autocapitalize: string|null|undefined;
  autoselect: boolean|null|undefined;
  clearButtonVisible: boolean;
  errorMessage: string;
  i18n: {clear: string};
  label: string;
  maxlength: number|null|undefined;
  minlength: number|null|undefined;
  name: string|null|undefined;
  placeholder: string|null|undefined;
  readonly: boolean|null|undefined;
  required: boolean|null|undefined;
  value: string;
  invalid: boolean;
  hasValue: boolean|null|undefined;
  preventInvalidInput: boolean|null|undefined;
  _enabledCharPattern: string|null|undefined;
  _createConstraintsObserver(): void;
  _onChange(e: Event): void;
  _valueChanged(newVal: unknown|null, oldVal: unknown|null): void;
  _constraintsChanged(required: boolean|null, minlength: number|null, maxlength: string|null, pattern: any): void;
  checkValidity(): boolean;
  ready(): void;
  validate(): boolean;
  clear(): void;
  _onKeyDown(e: KeyboardEvent): void;
  attributeChangedCallback(prop: string, oldVal: unknown|null, newVal: unknown|null): void;
}
```

### `vaadin-text-field.d.ts`

```ts
import {PolymerElement} from '@polymer/polymer/polymer-element.js';

import {TextFieldMixin} from './vaadin-text-field-mixin.js';

import {ControlStateMixin} from '@vaadin/vaadin-control-state-mixin/vaadin-control-state-mixin.js';

import {ThemableMixin} from '@vaadin/vaadin-themable-mixin/vaadin-themable-mixin.js';

import {ElementMixin} from '@vaadin/vaadin-element-mixin/vaadin-element-mixin.js';

import {html} from '@polymer/polymer/lib/utils/html-tag.js';

declare class TextFieldElement extends
  TextFieldMixin(
  ControlStateMixin(
  ThemableMixin(
  ElementMixin(
  PolymerElement)))) {
  list: string|null|undefined;
  pattern: string|null|undefined;
  title: string;
}

declare global {
  interface HTMLElementTagNameMap {
    "vaadin-text-field": TextFieldElement;
  }
}

export {TextFieldElement};
```


src/vaadin-number-field.html
```diff
@@ -325,29 +339,42 @@
...
+          this.setAttribute('step', newVal);
         }
 
+        /** @private */
         _minChanged(min) {
           this.inputElement.min = min;
         }
 
+        /** @private */
         _maxChanged(max) {
```
Haprog, line 369: I'm not sure why these are private when other property observers like `_stepChanged` and `_valueChanged` are protected?

src/vaadin-text-field-mixin.html
```diff
@@ -334,6 +341,10 @@
       ];
     }
 
+    /**
+     * @return {HTMLElement}
```
Haprog, line 345: Since there is:
```js
if (!this.shadowRoot) {
  return;
}
```
which returns `undefined`, maybe it should be:
```suggestion
     * @return {HTMLElement | undefined}
```
?

src/vaadin-text-field-mixin.html
```diff
@@ -345,17 +356,20 @@
       return this.shadowRoot.querySelector('[part="value"]');
     }
 
-    /**
-     * @private
-     */
+    /** @private */
     get inputElement() {
```
Haprog, line 360: I know this was already marked as private, but as it's used quite a lot in the other files, I'd consider making this protected. Not a blocker.

src/vaadin-text-field-mixin.html
```diff
@@ -523,6 +555,12 @@
       }
     }
 
+    /**
+     * @param {boolean?} required
+     * @param {number?} minlength
+     * @param {number?} maxlength
+     * @param {string?} maxlength
+     */
     _constraintsChanged(required, minlength, maxlength, pattern) {
```
Haprog, line 564: Looks like this is missing `@protected`? Also `?` suffix doesn't mean anything in JSDoc. `?` as a prefix would be nullable type and `=` as a suffix would be optional parameter.
```suggestion
    /**
     * @param {boolean | undefined} required
     * @param {number | undefined} minlength
     * @param {number | undefined} maxlength
     * @param {string | undefined} maxlength
     * @protected
     */
    _constraintsChanged(required, minlength, maxlength, pattern) {
```

Haprog: `=` suffix could also work here, but we'd need to also annotate the private override of `_constraintsChanged` in number-field with optional params since it adds a new param at the end and otherwise the generator complains that you can't have a non-optional parameter after optional params.

src/vaadin-text-field-mixin.html
```diff
@@ -776,11 +835,15 @@
...
       this._dispatchIronResizeEventIfNeeded('Height', this.offsetHeight);
     }
 
     /**
+     * @param {string} prop
+     * @param {unknown} oldVal
+     * @param {unknown} newVal
      * @protected
      */
     attributeChangedCallback(prop, oldVal, newVal) {
```
Haprog, line 849: `oldVal` and `newVal` types could also be string since these are attribute values which can only be strings (or also undefined in this case I think). Should also maybe add the missing `namespace` param that is documented on Polymer even if we don't use it.

See https://polymer-library.polymer-project.org/3.0/api/polymer-element#PolymerElement-method-attributeChangedCallback

src/vaadin-number-field.html
web-padawan: Those observers that are overridden by `vaadin-integer-field` are marked as protected.

I would prefer to keep others as private unless we need to override them as well.

src/vaadin-text-field-mixin.html
web-padawan: Thanks, changed.

src/vaadin-text-field-mixin.html
web-padawan: Done.

src/vaadin-text-field-mixin.html
web-padawan: Done.

src/vaadin-text-field-mixin.html
web-padawan: Done.

src/vaadin-number-field.html
```diff
@@ -325,29 +339,42 @@
...
         _allowed(sign) {
           const incr = sign * (this.step || 1);
           const value = parseFloat(this.value);
           return !this.value || (!this.disabled && this._incrementIsInsideTheLimits(incr, value));
         }
 
-        _stepChanged(step) {
+        /**
+         * @param {!number} newVal
+         * @param {!number} oldVal
```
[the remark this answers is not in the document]
web-padawan: Done.

src/vaadin-integer-field.html
```diff
@@ -71,6 +71,11 @@
           super._valueChanged(newVal, oldVal);
         }
 
+        /**
+         * @param {!number} newVal
+         * @param {!number} oldVal
```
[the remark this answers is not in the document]
web-padawan: Done.

web-padawan:
> property autoselect should have @type {boolean}

Done.

> enum types for autocomplete, autocorrect and autocapitalize

Added the enums:
- `autocomplete` based on the [W3C spec](https://www.w3.org/TR/html52/sec-forms.html#autofilling-form-controls-the-autocomplete-attribute)
- `autocapitalize` based on the [WHATWG spec](https://html.spec.whatwg.org/multipage/interaction.html#autocapitalization)
- `autocorrect` is not yet in the spec (https://github.com/whatwg/html/issues/3595), but I added it too.

BTW I looked into `lib.dom.d.ts` for `HTMLInputElement` and it only contains `autocomplete`:
```ts
    /**
     * Specifies whether autocomplete is applied to an editable text field.
     */
    autocomplete: string;
```
The two others are not supported by all browsers at the moment, which is probably the reason why there are no built-in TS definitions for them. But I think using enums is a good idea.

src/vaadin-text-field-mixin.html
Haprog: As this is now `@protected` it should have the `@return` type too.

@types/interfaces.d.ts
```diff
@@ -0,0 +1,54 @@
+export type TextFieldAutoCapitalize =
+  | 'on'
+  | 'off'
+  | 'none'
+  | 'characters'
+  | 'words'
+  | 'sentences';
+
+export type TextFieldAutoComplete =
```
Haprog, line 9: After looking at the spec you linked, I realized that `autocomplete` can't be a simple enum since it can have a "set of space-separated tokens" and also allows custom `section-*` token so it's practically impossible to simply list all the allowed values.

Therefore I think this should be changed back to just `{string}` so remove type `TextFieldAutoComplete`.

web-padawan: Done. Removed `@type` from the `autocomplete` property for now.

src/vaadin-text-field-mixin.html
web-padawan: Thanks, added.

Issue #475, opened 2020-05-22
Restore helper text slot and API


tomivirkki:
The feature was reverted some time ago. Restore (both Flow and Web component) and include in the next minor.

DiegoCardoso:
Fixed by #476

PR #476, opened 2020-05-25
Add helper text API and slot (#410)

DiegoCardoso:
Cherry-pick from the PR #410 that was reverted by #464.

Part of #475


src/vaadin-text-field-mixin.html
```diff
@@ -225,6 +226,15 @@
           observer: '_labelChanged'
         },
 
+        /**
+         * String used for the helper text.
```
web-padawan, line 230: Please add `@type {string | null}` for TS definitions, as we consider `null` supported.

src/vaadin-text-field-mixin.html
```diff
@@ -482,10 +498,28 @@
...
+    _helperTextChanged(helperText) {
+      this._setOrToggleAttribute('has-helper', helperText === 0 || !!helperText, this);
+    }
+
+    _errorMessageChanged(errorMessage) {
+      this._setOrToggleAttribute('has-error-message', errorMessage === 0 || !!errorMessage, this);
+    }
+
+    _onHelperSlotChange() {
+      const slottedNodes = this.root.querySelector(`[name="helper"]`).assignedNodes();
```
web-padawan, line 513: Please use `this.shadowRoot` instead of `this.root` which is the Polymer legacy API.

src/vaadin-text-field-mixin.html
```diff
@@ -642,6 +676,9 @@
       this.shadowRoot.querySelector('[name="input"], [name="textarea"]')
         .addEventListener('slotchange', this._onSlotChange.bind(this));
 
+      this._onHelperSlotChange();
+      this.shadowRoot.querySelector('[name="helper"]').addEventListener('slotchange', this._onHelperSlotChange.bind(this));
```
web-padawan, line 680: BTW we can use `FlattenedNodesObserver` so that we don't need to trigger `this._onHelperSlotChange()` here. Not a blocker.

src/vaadin-text-field-mixin.html
```diff
@@ -808,11 +846,15 @@
...
-        this.focusElement);
+    _getActiveErrorId(invalid, errorMessage, errorId, helperText, helperTextId, hasSlottedHelper) {
+      const ids = [];
+      if (helperText || hasSlottedHelper) {
+        ids.push(helperTextId);
+      }
+      if (errorMessage && invalid) {
+        ids.push(errorId);
+      }
+      this._setOrToggleAttribute('aria-describedby', ids.join(' '), this.focusElement);
```
web-padawan, line 857: Did we test this change with actual screen readers? If not, let's create a separate issue for it.

src/vaadin-text-field-mixin.html
```diff
@@ -482,10 +498,28 @@
...
     /** @private */
     _labelChanged(label) {
-      if (label !== '' && label != null) {
-        this.setAttribute('has-label', '');
-      } else {
-        this.removeAttribute('has-label');
+      this._setOrToggleAttribute('has-label', label === 0 || !!label, this);
+    }
+
+    _helperTextChanged(helperText) {
```
web-padawan, line 504: Please add `@private` annotation here.

src/vaadin-text-field-mixin.html
```diff
@@ -482,10 +498,28 @@
...
+      this._setOrToggleAttribute('has-helper', helperText === 0 || !!helperText, this);
+    }
+
+    _errorMessageChanged(errorMessage) {
```
web-padawan, line 508: Please add `@private` annotation here.

src/vaadin-text-field-mixin.html
```diff
@@ -482,10 +498,28 @@
...
+      this._setOrToggleAttribute('has-error-message', errorMessage === 0 || !!errorMessage, this);
+    }
+
+    _onHelperSlotChange() {
```
web-padawan, line 512: Please add `@private` annotation here.

src/vaadin-text-field-mixin.html
```diff
@@ -808,11 +846,15 @@
       }
     }
 
-    /** @private */
-    _getActiveErrorId(invalid, errorMessage, errorId) {
-      this._setOrToggleAttribute('aria-describedby',
-        (errorMessage && invalid ? errorId : undefined),
-        this.focusElement);
+    _getActiveErrorId(invalid, errorMessage, errorId, helperText, helperTextId, hasSlottedHelper) {
```
web-padawan, line 849: Please restore `@private` annotation here.

src/vaadin-text-field-mixin.html
DiegoCardoso: I am not sure about it. I gave a quick test here, and it looks like it's effective.

src/vaadin-text-field-mixin.html
DiegoCardoso: Changed.

src/vaadin-text-field-mixin.html
DiegoCardoso: Added.

src/vaadin-text-field-mixin.html
DiegoCardoso: Done.

src/vaadin-text-field-mixin.html
DiegoCardoso: Done.

src/vaadin-text-field-mixin.html
DiegoCardoso: Done.

src/vaadin-text-field-mixin.html
DiegoCardoso: Done.

PR #477, opened 2020-06-01
fix: use flattened nodes on helper slot

DiegoCardoso:
A 3rd custom element using `vaadin-text-field` and defining a slot area for the `vaadin-text-field`'s helper slot was making `vaadin-text-field` see the slot element as a slotted contend and setting `has-helper="slotted"` to itself.

This change uses flattened nodes to check the presence of slotted elements on `vaadin-text-field` `helper` slot.

In this change
- use `FlattenedNodesObserver` instead of `slotchange` event
- use `FlattenedNodesObserver.getFlattenedNodes` to get slot children
- apply changes to tests
- remove checks for unsupported types

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -6,6 +6,7 @@ This program is available under Apache License Version 2.0, available at https:/
 
 <link rel="import" href="../../polymer/lib/utils/async.html">
 <link rel="import" href="../../polymer/lib/utils/debounce.html">
+<link rel="import" href="../../polymer/lib/utils/flattened-nodes-observer.html">
 
 <dom-module id="vaadin-text-field-shared-styles">
   <template>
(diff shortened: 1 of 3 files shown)


tomivirkki:
Would using `.assignedNodes({flatten: true})` [here](https://github.com/vaadin/vaadin-text-field/blob/1f5b7b11c7f94c752a314547426867938ea58a90/src/vaadin-text-field-mixin.html#L517) fix the issue?

DiegoCardoso:
Hmm.. It looks it does indeed. Should I rollback the change to add the `FlattenedNodesObserver` for triggering this callback? Or keep it and use your suggestion? (that way I could still use `info.target` and avoid the query selector there)

tomivirkki:
Switching from native APIs towards Polymer helpers isn't recommended as it adds overhead to the Lit conversion where we need to use the native APIs anyway.

DiegoCardoso:
Reverted some of the changes to make it use native API and use the suggestion for getting the assigned nodes flattened.

PR #480, opened 2020-06-11
Fix user input

yuriy-fix:
Fixes https://github.com/vaadin/vaadin-text-field-flow/issues/273

**Description:**
`paste` is also triggering `input` event.
In case this event is triggered when value wasn't changed (i.e. value was replaced with `paste`, or `maxlength` is set and `paste` is done), `__userInput` will be set to `true` [here](https://github.com/vaadin/vaadin-text-field/blob/bd388d6bab1a08ae9fb5f68e95d513004c9592cb/src/vaadin-text-field-mixin.html#L428), but [updating the value](https://github.com/vaadin/vaadin-text-field/blob/bd388d6bab1a08ae9fb5f68e95d513004c9592cb/src/vaadin-text-field-mixin.html#L430) won't pass polymer dirty check, so the `__userInput` will be preserved for the next programmatic value update and input `value` won't be [updated](https://github.com/vaadin/vaadin-text-field/blob/bd388d6bab1a08ae9fb5f68e95d513004c9592cb/src/vaadin-text-field-mixin.html#L487).

**Important note:**
It's possible to check for specifically `paste` action (i.e. with `e.inputType` or within `onPaste`), but it doesn't make sense to preserve `__userInput` in any case if dirty check won't be passed for `valueChange` afterwards, so general short solution is implemented in the PR.
(It's easier to test and covers the `paste` cases, cannot come up with other cases when value won't be changed and `input` event is triggered)

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -424,11 +424,12 @@ This program is available under Apache License Version 2.0, available at https:/
         }
       }
 
-      // Value can remain the same on input when pasting
-      if (!e.__fromClearButton && this.value !== e.target.value) {
+      if (!e.__fromClearButton) {
         this.__userInput = true;
       }
+
       this.value = e.target.value;
+      this.__userInput = false;
     }
 
     // NOTE(yuriy): Workaround needed for IE11 and Edge for proper displaying
@@ -485,7 +486,6 @@ This program is available under Apache License Version 2.0, available at https:/
       }
 
       if (this.__userInput) {
-        this.__userInput = false;
         return;
       } else if (newVal !== undefined) {
         this.inputElement.value = newVal;
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -424,7 +424,8 @@ This program is available under Apache License Version 2.0, available at https:/
         }
       }
 
-      if (!e.__fromClearButton) {
+      // Value can remain the same on input when pasting
+      if (!e.__fromClearButton && this.value !== e.target.value) {
         this.__userInput = true;
       }
       this.value = e.target.value;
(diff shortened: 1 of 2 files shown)


src/vaadin-text-field-mixin.html
```diff
@@ -424,7 +424,8 @@
         }
       }
 
-      if (!e.__fromClearButton) {
+      // Value can remain the same on input when pasting
+      if (!e.__fromClearButton && this.value !== e.target.value) {
```
tomivirkki, line 428: Alternative fix would be to set `this.__userInput` back to `false` after `this.value` is set on line 430. Then we wouldn't have to rely on some side effect code somewhere to fix the flag back to `false`. wdyt?

yuriy-fix: Currently there is no need in setting the flag back to `false` as it won't be set to `true` incorrectly in the first place.
So I am avoiding one more place of removing the `userInput` flag this way.

But I don't mind other solution if it seems to be cleaner and I can use that one if needed :)

tomivirkki: > So I am avoiding one more place of removing the userInput flag this way

The place where it's currently set `false` could be removed, so it wouldn't add new logic but streamlines the existing logic.

The current removal of the flag is done somewhere else that relies on a side-effect of another action to trigger and that caused the issue in the first place (the observer didn't invoke if the set value happened to be the same as before).

IMO, toggling a flag like this would ideally be done in only one place. Now the logic is scattered so it's hard to follow and subject to errors.

yuriy-fix: Thanks for the clarification! That makes sense indeed as scattered logic was the cause of the issue I guess 😅 . 
Didn't consider moving removal from `_valueChanged`. Updated now.

PR #484, opened 2020-06-22
fix: workaround Polymer getProperty with mixins/observers

tomivirkki:
See https://github.com/Polymer/polymer/issues/5259

Fixes https://github.com/vaadin/vaadin-text-field-flow/issues/317

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -923,6 +923,11 @@ This program is available under Apache License Version 2.0, available at https:/
       }
     }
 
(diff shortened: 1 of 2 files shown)


tomivirkki:
The suggested fix no longer works with the latest Polymer version because `_getProperty` was inlined [here](https://github.com/Polymer/polymer/commit/280f4f0a374c8dc36048fca95f05d9860f7f27a3#diff-9aa9463b45002c1971afac3d19e7d1c3L163-R164)

Changed the fix so that a getter/setter pair is defined for `__data` instead 😬🙈

PR #495, opened 2020-08-31
perf: Debounce offset height measure

### LINKED ISSUE #405: Rendering of many TextFields is slow
### Description
If you have a form with many `TextField`s (`NumberField`s, `DatePicker`s etc. work too), the rendering becomes pretty slow.
The reason for this seems to be a so called [Forced Reflow](https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing), caused by TextFieldMixin#__observeOffsetHeight.

### Expected outcome
The method `__observeOffsetHeight` should not cause a poor rendering performance, accessing the `offsetHeight` should not trigger a forced reflow.

### Actual outcome
Accessing the `offsetHeight` inside `__observeOffsetHeight` causes a forced reflow and slows down the rendering.

### Steps to reproduce
1. Download the [Vaadin Starter](https://vaadin.com/start/pre-release) (I used the `Plain Java Servlet`)
2. Add the following code to `MainView.java`:
```Java
    private Button btn;
    private VerticalLayout form;
    private HashMap<String, String> model;
    public MainView() {
        model = new HashMap<>();
        setSizeFull();
        btn = new Button("Create Form", e -> recreateForm());
        btn.getStyle().set("cursor", "pointer");
        add(btn);
    }
    private void recreateForm() {
        if (form != null)
            remove(form);
        generateForm();
        add(form);
    }
    private void generateForm() {
        Binder<HashMap<String, String>> binder = new Binder<>();
        form = new VerticalLayout();
        VerticalLayout group = null;
        HorizontalLayout hGroup = null;
        for (int i = 0; i < 500; i ++) {
            if (group == null || i % 5 == 0)
                form.add(group = new VerticalLayout());
            if (hGroup == null || i % 2 == 0)
                group.add(hGroup = new HorizontalLayout());
            long rand = Math.round(Math.random() * 100) * (i + 1);
            String id = "tf" + i;
            TextField tf = new TextField("tf" + rand);
            binder.bind(tf, map -> map.get(id), (map, value) -> map.put(id, value));
            model.put(id, "v" + rand);
            hGroup.add(tf);
        }
        binder.readBean(model);
        binder.addValueChangeListener(e -> {
            try {
                binder.writeBean(model);
            }
            catch (ValidationException exc) {
            }
        });
    }
```

3. Open the Browser and click the `Create Form` Button.
   The rendering took about 4s in my case.
   Removing the `this._dispatchIronResizeEventIfNeeded('Height', this.offsetHeight);` from the `__observeOffsetHeight` method reduced the rendering time to about 2.5s.

### Screenshots
Chrome Performance analysis without patching the `TextFieldMixin`:
image

Bottom-Up Tab of the performance analysis:
image

Chrome Performance analysis after patching the `TextFieldMixin`:
image

### Notes
Executing `this._dispatchIronResizeEventIfNeeded('Height', this.offsetHeight);` in `beforeNextRender` achieved similar performance as removing it completely.
As I am pretty new to Polymer, I don't know if it has any side effects, but in my case the result looked the same.

### Browsers Affected
I tested it in Chrome

tomivirkki:
fixes #405 

This PR debounces the currently synchronous offsetHeight measurement (which causes a reflow) inside `__observeOffsetHeight` so that it's done on animation frame instead. The change has a significant impact on https://github.com/vaadin/flow/issues/8853 (page render time drops approximately 40%)


web-padawan (approved):
LGTM. Should also fix #405 which appears to be the same issue as mentioned above.

tomivirkki:
> Should also fix #405 which appears to be the same issue as mentioned above.

👍  Connected the issue.

PR #569, opened 2021-02-08
fix: scope a Safari-specific workaround

tomivirkki:
Cherry-pick https://github.com/vaadin/vaadin-text-field/pull/568

### DIFF
--- a/src/vaadin-text-area.html
+++ b/src/vaadin-text-area.html
@@ -200,8 +200,6 @@ This program is available under Apache License Version 2.0, available at https:/
             // Fix the input element width so its scroll height isn't affected by host's disappearing scrollbars
             input.style.maxWidth = inputWidth;
             input.style.height = 'auto';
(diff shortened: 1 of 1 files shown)


web-padawan (approved):

tomivirkki:
@web-padawan 

Seems that the workaround is still needed for the P2-codebase version. The tests were failing on:
- P2: Safari Desktop & iOS 12 simulator
- P3 (transpiled): iOS 12 simulator

Neither affects the next-gen version so it should be fine to only have the workaround in the P2 -codebase version.

web-padawan:
@tomivirkki good to know, thanks. I think it's ok to keep the workarounds only for 2.x

web-padawan (approved):

tomivirkki:
Seems that we might need to scope this in 3.x also :/ https://github.com/vaadin/vaadin-text-field/runs/1869591236

PR #570, opened 2021-02-10
fix: iron-resize event is now composed

sosa-vaadin:
Fixes https://github.com/vaadin/vaadin-date-picker/issues/788

### DIFF
--- a/src/vaadin-text-field-mixin.js
+++ b/src/vaadin-text-field-mixin.js
@@ -827,7 +827,7 @@ export const TextFieldMixin = (subclass) =>
     _dispatchIronResizeEventIfNeeded(sizePropertyName, value) {
       const previousSizePropertyName = '__previous' + sizePropertyName;
       if (this[previousSizePropertyName] !== undefined && this[previousSizePropertyName] !== value) {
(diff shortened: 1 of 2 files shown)


tomivirkki (approved):

web-padawan (commented):
While the workaround should solve this particular problem, it's against the principles of `IronResizableBehavior` where the [iron-resize event is non-bubbling](https://github.com/PolymerElements/iron-resizable-behavior/blob/4e3ede30d8c5175bd315250a88f85e4479453e77/iron-resizable-behavior.js#L32).

Instead we should apply the behavior and then fire `iron-request-resize-notifications` event, like [vaadin-board-row](https://github.com/vaadin/vaadin-board/blob/362b5b7e3a29c0b2be8fb719b91e3c7d45089f67/src/vaadin-board-row.js#L98-L106) does. The `vaadin-grid` would then use that event to assign itself as a parent.

tomivirkki:
> Instead we should apply the behavior

`iron-request-resize-notifications` isn't composed so to fix this "properly" we'd need to add the behavior to all components that use a text-field internally?

web-padawan:
@tomivirkki makes sense, thanks. So let's go with a current PR and reconsider the changes later.

PR #573, opened 2021-02-13
fix: 'iron-resize' event when height changes is now composed.

sosa-vaadin:


test/text-field.html
```diff
@@ -740,6 +740,18 @@
             flushTextField(textField);
           });
 
+          it('should be a composed event', done => {
+            setTimeout(() => {
```
tomivirkki, line 744: If flushing alone doesn't cut it (like in the P3-version test), would an event listener (for "iron-resize") work here instead? Timeouts in tests are generally unreliable and slow.

sosa-vaadin: An event listener works, so I will change it. We should probably change other tests that use timeouts then.

tomivirkki: If a callback is used (a timeout or an event listener), then the test must be asynchronous. Now that the `done` parameter was removed, the test is no longer async and will pass even without the callback (or with something like `expect(false).to.be.true;` in the callback).

tomivirkki: Since we're only interested in analyzing the event, and it seems that the [very last test](https://github.com/vaadin/vaadin-text-field/blob/0ce7937643cb7bee0e52f921da96e9def7a0abd2/test/text-field.html#L764-L767) in the section works synchronously, we could use the same sequence for this test also to avoid adding any callbacks.

tomivirkki (approved):

PR #578, opened 2021-06-23
refactor: use overlapping grid elements to grow/shrink textarea's height

fhdhsni:
This PR fixes the issue with scroll jumping to top of the page on certain circumstances. It solves it with help of `grid` in css and a `div` and a `textarea` on top of each other so the change in height of the `div` is reflected on `textarea`.

Internet explorer does not support `grid` so we used another approach for this case, other browsers use the correct approach.

There was also an issue with running tests on iOS 10.3 (that the height of the textarea was not calculated correctly, it didn't inherit the height of its parent), in order to fix this, we checked the height of the wrapper div instead of the textarea itself 

There was also a slight difference in one of the visual tests about material theme, I updated the screenshots and there is a very slight change in size of scroll bar which is barely visible (an no change in box-model of component).

Here is rest of Farhad's comment:

shady css compiles:

`foo ::slotted(bar)`

to:

`foo > bar`

which results in breaking some of the styles with the addition of
`.textarea-wrapper`. To prevent that I had to copy and paste some of the
styles from vaadin-text-field-mixin.html to vaadin-text-area.html and
change the selector from `[part="input-field"] ::slotted(textarea)` to
`.textarea-wrapper ::slotted(textarea)`


yuriy-fix (commented):
There is minor visual regression on IE11 (text-color, size). 
Comparing `2.8` and `refactor/text-area-auto-height` branches:

Capture-2
Capture

src/vaadin-text-area.html
```diff
@@ -189,18 +266,37 @@
...
+          const inputField = this.root.querySelector('[part=input-field]');
+          const scrollTop = inputField.scrollTop;
+          const inputWidth = getComputedStyle(input).width;
+          const y = window.pageYOffset || document.documentElement.scrollTop;
           // Only clear the height when the content shortens to minimize scrollbar flickering.
           if (this._oldValueLength >= valueLength) {
             // Fix the input element width so its scroll height isn't affected by host's disappearing scrollbars
             input.style.maxWidth = inputWidth;
             input.style.height = 'auto';
+            // input.style.padding = '0';
```
alvarezguille, line 299: Please remove commented code

src/vaadin-text-area.html
```diff
@@ -215,8 +311,10 @@
 
           // Restore
           input.style.removeProperty('max-width');
+          // input.style.removeProperty('padding');
```
alvarezguille, line 314: Please remove commented code

src/vaadin-text-area.html
```diff
@@ -27,17 +27,92 @@
...
+          resize: none;
+          overflow: hidden;
+        }
+
+        [part='value'],
+        .textarea-wrapper ::slotted(textarea),
+        .textarea-wrapper::after {
+          grid-area: 1 / 1 / 2 / 2;
+          box-sizing: border-box;
+          /* padding: 0 0.25em; */
```
alvarezguille, line 110: Please remove commented code

src/vaadin-text-area.html
```diff
@@ -27,17 +27,92 @@
...
+        line-height: normal;
+        background-color: transparent;
+        box-shadow: none;
+
+        flex: auto;
+        overflow: hidden;
+        width: 100%;
+        height: 100%;
+        padding-bottom: calc((var(--lumo-text-field-size) - 1em * var(--lumo-line-height-s)) / 2);
+        /* color: var(--lumo-body-text-color); */
```
alvarezguille, line 74: Please remove commented code

src/vaadin-text-area.html
```diff
@@ -27,17 +27,92 @@
...
+        flex: auto;
+        box-sizing: border-box;
+        height: 100%;
+      }
+
       [part="value"],
-      [part="input-field"] ::slotted(*) {
-        align-self: flex-start;
+      .textarea-wrapper ::slotted(textarea) {
+        /* reset ua stylesheet */
```
alvarezguille, line 54: what's `ua` in this comment?

alvarezguille (changes requested):
Added some initial contents, I'll continue testing locally

alvarezguille (changes requested):
I'm getting this behavior on initial load of the demos in ios safari 13.1, it happens only on the initial load, and get's fixed after interacting
image
@fhdhsni @hdamr could you check if the issue is related to the change?

sissbruecker:
Obsolete due to #581

PR #581, opened 2021-09-01
fix: prevent page scroll when resizing text area

sissbruecker:
## Description

This change prevents the page from scrolling while the text area is measuring its height as part of the auto resize feature.

The solution involves temporarily fixing the height of the `textarea` container element, while the height of the `textarea` element is set to `auto` for measuring purposes. By setting a fixed height on the container we can ensure that the page height does not change during measurement, which prevents the browser from automatically adjusting the scroll position of the page.

Fixes https://github.com/vaadin/web-components/issues/291

Merging this would make #578 obsolete.

## Type of change

- [x] Bugfix
- [ ] Feature

### DIFF
--- a/src/vaadin-text-area.html
+++ b/src/vaadin-text-area.html
@@ -164,7 +164,6 @@ This program is available under Apache License Version 2.0, available at https:/
           super.ready();
           this._updateHeight();
           this.addEventListener('animationend', this._onAnimationEnd);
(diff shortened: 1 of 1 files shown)


sissbruecker:
@alvarezguille Couldn't find anything that would prevent that behaviour in FF. Searching the web there are some hacks for forcing the browser to scroll to the cursor / caret, but nothing that looks reliable or wouldn't lead to other issues.

alvarezguille (approved):

PR #582, opened 2021-09-27
Added pattern validation to text area

roastedcpu:
## Description

Adds pattern validation to text area; supports multiline validation
Motivation: https://github.com/vaadin/flow-components/issues/2077 (Support request VS-3922)
Acceptance criteria: https://github.com/vaadin/platform/issues/2377

Part of: https://github.com/vaadin/flow-components/issues/2077
AC: https://github.com/vaadin/platform/issues/2377

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [x] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

### DIFF
--- a/test/text-area.html
+++ b/test/text-area.html
@@ -110,6 +110,13 @@
               assertAttrCanBeSet(prop, false);
             });
           });
+
+          ['pattern'].forEach(function(prop) {
+            it('should set string attribute ' + prop, function() {
+              assertAttrCanBeSet(prop, 'none');
+            });
+          });
+
         });
 
         describe(`binding ${condition}`, () => {
--- a/demo/text-area-demos.html
+++ b/demo/text-area-demos.html
@@ -48,6 +48,18 @@
       </template>
     </vaadin-demo-snippet>
 
+    <h3>Pattern</h3>
+    <vaadin-demo-snippet id="text-area-demos-text-area-with-pattern">
+      <template preserve-content>
+        <vaadin-text-area
+            class="pattern"
+            label="Description"
+            pattern="[0-9.,]+"
+            placeholder="Write here ...">
+        </vaadin-text-area>
+      </template>
+    </vaadin-demo-snippet>
+
   </template>
   <script>
     class TextAreaDemos extends DemoReadyEventEmitter(TextFieldDemo(Polymer.Element)) {
--- a/src/vaadin-text-area.html
+++ b/src/vaadin-text-area.html
@@ -233,6 +233,19 @@ This program is available under Apache License Version 2.0, available at https:/
           this._dispatchIronResizeEventIfNeeded('InputHeight', inputHeight);
         }
 
+        checkValidity() {
+          if (!super.checkValidity()) {
+            return false;
+          }
+
+          if (!this.pattern) {
+            return true;
+          }
+
+          const match = this.inputElement.value.match(this.pattern);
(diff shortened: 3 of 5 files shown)


test/text-area.html
```diff
@@ -110,6 +110,19 @@
...
+              assertAttrCanBeSet(prop, 'none');
+            });
+          });
+
+          ['multilinePatternMode'].forEach(function(prop) {
+            it('should set boolean attribute ' + prop, function() {
+              assertAttrCanBeSet(prop, true);
+              assertAttrCanBeSet(prop, false);
+            });
+          });
```
sissbruecker, line 125: There are lots of missing tests:
- test that textarea becomes invalid if input does not match pattern
- test that validation state updates when changing the pattern while the textarea has an existing value
- test that value does not change when setting preventInvalidInput and inputting an invalid character
- test that multiline pattern mode works as expected
- test that validation state updates when changing multiline pattern while textarea has an existing value

sissbruecker (changes requested):
In general there seem to be no acceptance criteria defined for the feature, which is usually required to accept a feature contribution.

For `pattern` and `preventInvalidInput` we can assume that they should work equivalent to the text field. However `multilinePatternMode` seems very open to interpretation and it should be specified how it should work, or if we want to include it at all.

Please confer with the PO (@rolfsmeds ) to define acceptance critera for the feature.

rolfsmeds:
> In general there seem to be no acceptance criteria defined for the feature, which is usually required to accept a feature contribution. See this part of the checklist for PRs:
> 
> > Additional for Feature type of change
> > Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.
> 
> For `pattern` and `preventInvalidInput` we can assume that they should work equivalent to the text field. However `multilinePatternMode` seems very open to interpretation and it should be specified how it should work, or if we want to include it at all.
> 
> Please confer with the PO (@rolfsmeds ) to define acceptance critera for the feature.

AC is here actually: https://github.com/vaadin/platform/issues/2377

sissbruecker:
According to the acceptance criteria:
- there is no mention of the multiline pattern mode. So we either need to define ACs for that as well, or we remove it
- there needs to be a test that verifies that our custom regex pattern validation is compliant with the web standards, by making it pass with these test cases: https://github.com/web-platform-tests/wpt/blob/7b0ebaccc62b566a1965396e5be7bb2bc06f841f/html/semantics/forms/constraints/form-validation-validity-patternMismatch.html

test/text-area.html
roastedcpu: I think those tests are already there for text area as well, under test/validation.html

roastedcpu:
@sissbruecker It seems that some tests are still missing, according to the AC. Please do not merge this before I add them :) I think all your points have been addressed by now. Thank you!

sissbruecker:
Closing in favor of #586

PR #584, opened 2021-11-05
fix: workaround disabled textarea freezing Firefox (#2981)

tomivirkki:
Backport https://github.com/vaadin/web-components/pull/2981 for V14


vursen (approved):
Confirmed that the bug is no longer reproducible in V14 with the proposed fix (I could reproduce it without the fix).

https://user-images.githubusercontent.com/5039436/140518123-01b1c14d-01a3-4987-b3f8-92ad0a57bdd0.mov

web-padawan (approved):

PR #585, opened 2021-11-08
refactor: make keydown event listener an arrow function to allow overwrites

sosa-vaadin:
## Description

The `vaadin-combo-box` component uses `vaadin-text-field` internally and needs to do some actions when the <kbd>Esc</kbd> key is pressed. These actions include changing the value of the text field, but there's an event listener in `vaadin-text-field-mixin` that also acts on its value.

To let `vaadin-combo-box` overrite this event listener, we must change the callback to an arrow function that calls the function we use and then override the function being called inside the arrow function in the particular instance in the combo-box.

Related https://github.com/vaadin/web-components/issues/984

## Type of change

- Refactor


tomivirkki:
I guess this can be closed (https://github.com/vaadin/vaadin-combo-box/pull/1015 no longer depends on this)?

sosa-vaadin:
> I guess this can be closed ([vaadin/vaadin-combo-box#1015](https://github.com/vaadin/vaadin-combo-box/pull/1015) no longer depends on this)?

As long as it can be merged without this, yes.
Closing.

PR #587, opened 2021-12-30
fix: Refactor vaadin-text-area Material styles

TatuLund:
fixes: https://github.com/vaadin/web-components/issues/1333 for Vaadin 14

Note1: Vaadin 22 has altered DOM structure and this fix cannot be applied there as is


TatuLund:
Just note. Some visual tests break as this change grows the undefined height field by 2px as the underline is growing "outside". However with these styles it looks good and field with fixed height looks good.

This screenshot breaks since the bug is present in refence shot
https://github.com/vaadin/vaadin-text-field/blob/master/test/visual/screens/vaadin-text-field/text-area-1-material/text-area-1/chrome.png

TatuLund:
Closing in favor of https://github.com/vaadin/web-components/pull/3359

PR #589, opened 2022-02-01
fix: position the material text-area bottom bar (CP)

tomivirkki:
Backport https://github.com/vaadin/web-components/pull/3359 for Vaadin 14


web-padawan (approved):

StainlessSteelRat:
This seems to break wheel scrolling a container which has a text-area inside it.  If the mouse pointer is over the text-area, then the container won't scroll.  I guess this is because the textarea is consuming the scroll event. (If the mouse pointer is over anywhere else on the container or other componet in the container, then scrolling works fine.)

tomivirkki:
Hi @StainlessSteelRat, yeah it's the `e.preventDefault()` in the new "wheel" event listener that causes it. We should change it so that the event gets canceled only if the text area's scroll position changes. Please report as a new issue in https://github.com/vaadin/web-components

PR #611, opened 2023-03-21
fix: prevent initial validation when the field is initialized as invalid (CP: 14)

roastedcpu:
Backports web-components#4138 as part of an EoD task

> ## Description
> The PR prevents the initial validation that could previously take place when the field is initially provided with a non-empty value and is initially marked `invalid`:
> 
> ```js
> const textField = document.createElement('vaadin-text-field');
> textField.value = 'Initial Value';
> textField.invalid = true;
> document.appendChild(textField);
> ```
> 
> The initial validation is problematic because it fires a `validated` event that may result in an `invalid` state reset on the Flow side, see [vaadin/flow-components#3429 (comment)](https://github.com/vaadin/flow-components/pull/3429#issuecomment-1177365424) for a more detailed case.
> 
> Part of #4150
> 
> ## Covered components
> This PR only fixes the issue for the components that extend `InputFieldMixin`:
> 
> * [x]  text-field
> * [x]  email-field
> * [x]  password-field
> * [x]  text-area
> * [x]  number-field
> * [x]  integer-field
> 
> Other components will be covered in the next PRs.
> 
> ## Type of change
> * [x]  Bugfix
> 
> ## Checklist
> * [x]  I have read the contribution guide: https://vaadin.com/docs-beta/latest/guide/contributing/overview/
> * [x]  I have added a description following the guideline.
> * [x]  The issue is created in the corresponding repository and I have referenced it.
> * [x]  I have added tests to ensure my change is effective and works as intended.
> * [x]  New and existing tests are passing locally with my change.
> * [x]  I have performed self-review and corrected misspellings.

### DIFF
--- a/test/validation.html
+++ b/test/validation.html
@@ -13,7 +13,7 @@
   <link rel="import" href="../vaadin-text-area.html">
   <link rel="import" href="../vaadin-text-field.html">
   <link rel="import" href="../../iron-form/iron-form.html">
-
+  <link rel="import" href="helpers.html">
 </head>
 
 <body>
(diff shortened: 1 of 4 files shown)


test/validation.html
```diff
@@ -334,7 +334,40 @@
             expect(field.invalid).to.be.true;
           });
         });
+      });
+    });
+
+    describe('initial validation', () => {
```
vursen, line 340: Let's create separate files per component – `text-field-validation.html`, `number-field-validation.html`, etc – with the following structure:

```js
describe('text-field validation', () => {
  describe('initial', () => {
    // add tests related to the initial validation here
  });

  describe('basic', () => {
    // add the validated event tests here
  });

  describe('invalid cannot be set to false', () => {
    // add tests related to shouldSetInvalid here
  });
});
```

Let's try that approach in this PR and then update other PRs after this one is merged.

PR #615, opened 2023-04-17
feat: add methods to make setting invalid overridable

mukherjeesudebi:
## Description

Backports the following PRs as part of an EoD task:

- https://github.com/vaadin/web-components/pull/4097

### DIFF
--- a/test/password-field-validation.html
+++ b/test/password-field-validation.html
@@ -17,6 +17,11 @@
       <vaadin-password-field></vaadin-password-field>
     </template>
   </test-fixture>
+  <test-fixture id="should-set-invalid">
+    <template>
+      <vaadin-password-field></vaadin-password-field>
+    </template>
+  </test-fixture>
 <script>
   describe(`initial validation`, () => {
     let field, validateSpy;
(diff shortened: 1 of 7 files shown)


test/password-field-validation.html
```diff
@@ -79,5 +84,23 @@
       expect(event.detail.valid).to.be.false;
     });
   });
+  describe('invalid cannot be set to false', () => {
+    let passwordField;
+
+    beforeEach(async() => {
+      passwordField = fixture('should-set-invalid');
```
vursen, line 91: I'm wondering if `password-field` would be a better naming for the `basic` fixture after all. It would allow us to re-use it here then.

```js
describe('basic', () => {
  beforeEach(() => {
    passwordField = fixture('password-field');
  });
});

describe('invalid cannot be set to false', () => {
  beforeEach(() => {
    passwordField = fixture('password-field');
  });
});
```

test/text-field-validation.html
```diff
@@ -79,5 +79,24 @@
...
   });
+  describe('invalid cannot be set to false', () => {
+    let textField;
+
+    beforeEach(async() => {
+      textField = fixture('text-field');
+      textField._shouldSetInvalid = (invalid) => invalid;
+      await nextRender();
+    });
+
```
vursen, line 90:
```suggestion
```

vursen (approved):

PR #614, opened 2023-04-05
feat: fire a validated event on validation (CP: 14)

mukherjeesudebi:
## Description

Backports the following PRs as part of an EoD task:

- https://github.com/vaadin/web-components/pull/4084

PR #616, opened 2023-04-18
docs: add JSDoc for validated event (CP: 14)

mukherjeesudebi:
## Description

This PR adds the JSDoc for the validated event that was added as a new feature 

- https://github.com/vaadin/vaadin-text-field/pull/614

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -986,11 +986,11 @@ This program is available under Apache License Version 2.0, available at https:/
      */
 
     /**
-      * Fired whenever the field is validated.
-      *
-      * @event validated
-      * @param {Object} detail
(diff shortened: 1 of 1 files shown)


src/vaadin-text-field-mixin.html
```diff
@@ -984,5 +984,13 @@
...
      * @event iron-resize
      */
+
+    /**
+      * Fired whenever the field is validated.
+      *
+      * @event validated
+      * @param {Object} detail
+      * @param {boolean} detail.valid the result of the validation.
+      */
```
vursen, line 994:
```suggestion
    /**
     * Fired whenever the field is validated.
     *
     * @event validated
     * @param {Object} detail
     * @param {boolean} detail.valid the result of the validation.
     */
```

vursen (approved):

PR #618, opened 2023-04-21
feat: add internal has-input-value-changed event

mukherjeesudebi:
## Description

Backports the following PRs:

- https://github.com/vaadin/web-components/pull/4276
- https://github.com/vaadin/web-components/pull/4335
- https://github.com/vaadin/web-components/pull/5092

Part of https://github.com/vaadin/web-components/issues/5763

PR #620, opened 2023-04-24
fix: check for bad input when updating _hasInputValue

mukherjeesudebi:
## Description

Backports the following PRs:

- https://github.com/vaadin/web-components/pull/4276
- https://github.com/vaadin/web-components/pull/4335
- https://github.com/vaadin/web-components/pull/5092

Part of  https://github.com/vaadin/vaadin-text-field/pull/618

### DIFF
--- a/src/vaadin-number-field.html
+++ b/src/vaadin-number-field.html
@@ -418,6 +418,22 @@ This program is available under Apache License Version 2.0, available at https:/
 
           return super.checkValidity();
         }
+
+        /**
+          * Native [type=number] inputs don't update their value
+          * when you are entering input that the browser is unable to parse
+          * e.g. "--5", hence we have to override this method from `InputMixin`
+          * so that, when value is empty, it would additionally check
+          * for bad input based on the native `validity.badInput` property.
+          *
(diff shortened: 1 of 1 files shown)


vursen (approved):

src/vaadin-number-field.html
```diff
@@ -418,6 +418,21 @@
...
+          * Native [type=number] inputs don't update their value
+          * when you are entering input that the browser is unable to parse
+          * e.g. "--5", hence we have to override this method from `InputMixin`
+          * so that, when value is empty, it would additionally check
+          * for bad input based on the native `validity.badInput` property.
+          *
+          * @param {InputEvent} event
+          * @protected
+          * @override
+          */
```
vursen, line 432:
```suggestion
         * Native [type=number] inputs don't update their value
         * when you are entering input that the browser is unable to parse
         * e.g. "--5", hence we have to override this method from `InputMixin`
         * so that, when value is empty, it would additionally check
         * for bad input based on the native `validity.badInput` property.
         *
         * @param {InputEvent} event
         * @protected
         * @override
         */
```

PR #621, opened 2023-04-25
refactor: input value getters and setters (CP: 14)

mukherjeesudebi:
## Description

Backports the following PRs:

- https://github.com/vaadin/web-components/pull/5586

### DIFF
--- a/src/vaadin-text-field-mixin.html
+++ b/src/vaadin-text-field-mixin.html
@@ -417,6 +417,39 @@ This program is available under Apache License Version 2.0, available at https:/
     }
 
     /**
+     * A property for accessing the input element's value.
+     *
+     * Override this getter if the property is different from the default `value` one.
+     *
+     * @protected
+     * @return {string}
+     */
+    get _inputElementValueProperty() {
+      return 'value';
+    }
+
+    /**
+     * The input element's value.
+     *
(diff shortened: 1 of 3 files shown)


vursen:
The new getter can be also used here:

https://github.com/vaadin/vaadin-text-field/blob/7e0346552e9febd22a4032bff81809d18078eedd/src/vaadin-number-field.html#L289

https://github.com/vaadin/vaadin-text-field/blob/7e0346552e9febd22a4032bff81809d18078eedd/src/vaadin-text-area.html#L290

vursen:
And also here:

https://github.com/vaadin/vaadin-text-field/blob/7e0346552e9febd22a4032bff81809d18078eedd/src/vaadin-text-area.html#L284

src/vaadin-text-field-mixin.html
```diff
@@ -792,6 +825,8 @@
 
     clear() {
       this.value = '';
+
+      this._inputElementValue = '';
```
vursen, line 829: It shouldn't be necessary for text fields because there is no situation when `this.value` would hold an empty string and `this._inputElementValue` would not. This line might make sense for time-picker or date-picker.

vursen (approved):

